PR #27, opened 2022-03-25
Fix fee calculation of NativeScript transaction

siegfried:


SebastienGllmt (approved):

SebastienGllmt:
nit: rename the comment in `count_needed_vkeys` from "minting keys" -> "native script keys"

nit: you can rename `get_mint_scripts` to `get_native_scripts`. It's fine to make breaking changes

siegfried:
Updated. Also, how about adding a `set_vkey_number()` to `TransactionBuilder`? For 1/10 multi-sig transaction, the fee is calculated with 10 keys whereas 1 key is enough. Users could use this method to set the count explicitly when it is possible to apply a minimum fee.

SebastienGllmt:
@siegfried you can't implement a `set_vkey_number` without also adding an equivalent function in the dApp connector spec, otherwise wallets won't know to shrink the tx fee using the same technique so the fee will be wrong

siegfried:
I can do that when I have time, maybe in another PR if you want to merge this.

SebastienGllmt (approved):
Merging for now. Other tasks can be done as follow-up

PR #29, opened 2022-03-28
Add witness builder

SebastienGllmt:
This is the same witness builder PR I made to serialization-lib that never got merged


SebastienGllmt:
Merged since this was reviewed previously when submitted to CSL

PR #30, opened 2022-03-28
Add as_ conversions to the Address struct

SebastienGllmt:
Currently you can convert between address types doing, for example, `RewardAddress.from_address(...)`

A design philosophy during the CSL days was there should only be one way to do this. This was to avoid duplication of work when converting Yoroi code to Yoroi Mobile code which required native bindings.

However, this limitation is no longer required so we should add useful helper functions like this


rooooooooob (approved):
LGTM

PR #31, opened 2022-03-28
Rename project

SebastienGllmt:
Note: this may conflict with open PRs because it renames the Flow file


doc/conf.py
```diff
@@ -21,9 +21,9 @@
 
 # -- Project information -----------------------------------------------------
 
-project = 'cardano-serialization-lib Documentation'
-copyright = '2020, EMURGO'
-author = 'EMURGO'
+project = 'cardano-multiplatform-lib Documentation'
+copyright = '2020, dcSpark'
```
nicarq, line 25: maybe we could do 2022 🤔

PR #32, opened 2022-03-31
Fix: nint deserialization below i64::MIN

SebastienGllmt:
Port of https://github.com/Emurgo/cardano-serialization-lib/pull/392


rooooooooob (approved):
It'd be reviewing my own PR but if you're fine with the code changes then it LGTM. It's a small change anyway and has test cases.

PR #33, opened 2022-03-31
PointerAddress support for >u32::MAX

SebastienGllmt:
This is a port of https://github.com/Emurgo/cardano-serialization-lib/pull/367


rooooooooob (approved):
I'll do a global change for TransactionIndex/CertificateIndex in a new PR since I think we'll find those on-chain in other places besides pointer addresses too.

PR #35, opened 2022-03-31
Dcspark/auto json to ts

SebastienGllmt:
This is a port of https://github.com/Emurgo/cardano-serialization-lib/pull/345

This one has some merge conflicts though. I did some cleanup so hopefully it shouldn't be too bad


rooooooooob (approved):
I fixed some issues where scripts/files were still referring to cardano-serialization-lib + removed some unneeded build dependencies I forgot to remove before from the main Cargo.toml.

PR #41, opened 2022-04-06
Clippy fixes part 1

SebastienGllmt:
Once we do the regeneration, we want to make sure the code generated conforms to both `clippy` and `cargo fmt` so that we don't have to disable these tools for the repo.

However, that means we also have to fix clippy errors on the non-generated part of the codebase as well. This PR makes progress towards this, but definitely doesn't fix all issues (I decided to stop here because otherwise the PR would get too big and it's probably already a bit too big)

### DIFF
--- a/rust/pkg/cardano_multiplatform_lib.js.flow
+++ b/rust/pkg/cardano_multiplatform_lib.js.flow
@@ -815,10 +815,10 @@ declare export class BigInt {
   as_int(): Int | void;
 
   /**
-   * @param {string} text
+   * @param {string} string
    * @returns {BigInt}
    */
-  static from_str(text: string): BigInt;
+  static from_str(string: string): BigInt;
 
   /**
    * @returns {string}
@@ -7134,6 +7134,11 @@ declare export class TransactionUnspentOutputs {
    * @returns {TransactionUnspentOutputs}
    */
   static new(): TransactionUnspentOutputs;
+
+  /**
+   * @returns {boolean}
+   */
+  is_empty(): boolean;
 
   /**
    * @returns {number}
--- a/rust/src/plutus.rs
+++ b/rust/src/plutus.rs
@@ -572,7 +572,7 @@ pub enum RedeemerTagKind {
 }
 
 #[wasm_bindgen]
-#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd, Hash, serde::Serialize, serde::Deserialize, JsonSchema)]
+#[derive(Copy, Clone, Debug, Eq, Ord, PartialEq, PartialOrd, Hash, serde::Serialize, serde::Deserialize, JsonSchema)]
 pub struct RedeemerTag(RedeemerTagKind);
 
 to_from_bytes!(RedeemerTag);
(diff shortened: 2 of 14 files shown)


rust/src/plutus.rs
```diff
@@ -572,7 +572,7 @@ pub enum RedeemerTagKind {
 }
 
 #[wasm_bindgen]
```
rooooooooob, line 574: A lot of `plutus.rs` was code-generated originally but it's juts one line so it should be fine. `Copy` is a bit weird since it doesn't mean much from wasm (you till get memory freeing when passed by value like we got with `BigNum` iirc) and this class won't exist in rust (since we can use enums directly without needing to wrap them for wasm restrictions) after the regeneration.

rust/src/utils.rs
```diff
@@ -163,13 +163,18 @@ impl Deserialize for TransactionUnspentOutput {
 }
 
 #[wasm_bindgen]
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Default)]
```
rooooooooob, line 166: just FYI we're getting rid of these Vec-wrapper types outside of the wasm bindings in general since they were only needed due to `wasm_bindgen` restrictions. Although this struct would still need to exist, we'll just be moving it to the wasm bindings crate.

rust/src/utils.rs
```diff
@@ -194,13 +199,22 @@ pub struct BigNum(u64);
...
+
 #[wasm_bindgen]
 impl BigNum {
     // Create a BigNum from a standard rust string representation
+    #[allow(clippy::should_implement_trait)]
     pub fn from_str(string: &str) -> Result<BigNum, JsError> {
-        string.parse::<u64>()
-            .map_err(|e| JsError::from_str(&format! {"{:?}", e}))
-            .map(BigNum)
+        // have to redefine so it's visible in WASM
```
rooooooooob, line 216: Won't we be getting rid of this struct in general and just be using `u64` when we drop flow support?

rooooooooob (approved):
LGTM now

rust/pkg/cardano_multiplatform_lib.js.flow
```diff
@@ -815,10 +815,10 @@ declare export class BigInt {
   as_int(): Int | void;
 
   /**
-   * @param {string} text
```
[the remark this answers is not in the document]
SebastienGllmt: We used string in other places so I was just unifying the identifiers. This wasn't a clippy thing.
However, you have a point that it's not great and we can unify all the places to use something other than string as the name

rust/src/utils.rs
SebastienGllmt: Yes, but I just wanted to silence the clippy warning anyway

rust/pkg/cardano_multiplatform_lib.js.flow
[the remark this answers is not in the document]
rooooooooob: @SebastienGllmt  Any suggestions? A lot of these are probably going to be involved in the regeneration (not this class though I hand-wrote it) so if we can settle on one now I can update cddl-codegen to use that.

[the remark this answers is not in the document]
SebastienGllmt: Nah, I think you're right that probably calling it anything other than "string" is fine

PR #44, opened 2022-04-06
Expose get address header

SebastienGllmt:
We used to treat the header as an internal implementation detail, but this would be useful to export for indexers

### DIFF
--- a/rust/pkg/cardano_multiplatform_lib.js.flow
+++ b/rust/pkg/cardano_multiplatform_lib.js.flow
@@ -349,9 +349,12 @@ declare export var StakeCredKind: {|
 |};
 
 /**
+ * Careful: this enum doesn't include the network ID part of the header
+ * ex: base address isn't 0b0000_0000 but instead 0b0000
+ * Use `header_matches_kind` if you don't want to implement the bitwise operators yourself
  */
 
-declare export var AddressHeader: {|
+declare export var AddressHeaderKind: {|
   +BasePaymentKeyStakeKey: 0, // 0
   +BasePaymentScriptStakeKey: 1, // 1
   +BasePaymentKeyStakeScript: 2, // 2
@@ -456,6 +459,13 @@ declare export class Address {
    * @returns {number}
    */
   header(): number;
+
+  /**
+   * @param {number} header
+   * @param {number} kind
+   * @returns {boolean}
+   */
+  static header_matches_kind(header: number, kind: number): boolean;
 
   /**
    * @returns {Uint8Array}
--- a/rust/pkg/cardano_multiplatform_lib.js.flow
+++ b/rust/pkg/cardano_multiplatform_lib.js.flow
@@ -409,6 +409,11 @@ declare export class Address {
   static from_json(json: string): Address;
 
   /**
+   * @returns {number}
+   */
+  header(): number;
+
+  /**
    * @returns {Uint8Array}
    */
   to_bytes(): Uint8Array;
--- a/rust/pkg/cardano_multiplatform_lib.js.flow
+++ b/rust/pkg/cardano_multiplatform_lib.js.flow
@@ -351,6 +351,23 @@ declare export var StakeCredKind: {|
 /**
  */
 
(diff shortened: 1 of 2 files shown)


rust/src/address.rs
```diff
@@ -307,39 +307,45 @@ impl JsonSchema for Address {
...
+            AddrType::Ptr(ptr) => 0b0100_0000
+                               | ((ptr.payment.kind() as u8) << 4)
+                               | (ptr.network & 0xF),
+            AddrType::Enterprise(enterprise) => 0b0110_0000
+                               | ((enterprise.payment.kind() as u8) << 4)
+                               | (enterprise.network & 0xF),
+            AddrType::Reward(reward) => 0b1110_0000
+                                | ((reward.payment.kind() as u8) << 4)
+                                | (reward.network & 0xF),
+            AddrType::Byron(_) => 0b1000,
```
rooooooooob, line 324: 0b1000 are bits 7-4 so this should at least be 0b1000_0000. The explanation in `alonzo.cddl` doesn't mention network ID for byron addresses but they're the lower bits in other address types. Can you not use them outside of mainnet or was that just an omission in that header documentation? Do they map the old protocol magic to Shelley+ equivalents?

```
; shelley payment addresses:
; bit 7: 0
; bit 6: base/other
; bit 5: pointer/enterprise [for base: stake cred is keyhash/scripthash]
; bit 4: payment cred is keyhash/scripthash
; bits 3-0: network id
;
; reward addresses:
; bits 7-5: 111
; bit 4: credential is keyhash/scripthash
; bits 3-0: network id
;
; byron addresses:
; bits 7-4: 1000

; 0000: base address: keyhash28,keyhash28
; 0001: base address: scripthash28,keyhash28
; 0010: base address: keyhash28,scripthash28
; 0011: base address: scripthash28,scripthash28
; 0100: pointer address: keyhash28, 3 variable length uint
; 0101: pointer address: scripthash28, 3 variable length uint
; 0110: enterprise address: keyhash28
; 0111: enterprise address: scripthash28
; 1000: byron address
; 1110: reward account: keyhash28
; 1111: reward account: scripthash28
; 1001 - 1101: future formats
```

rooooooooob (changes requested):

rust/pkg/cardano_multiplatform_lib.js.flow
```diff
@@ -348,6 +348,23 @@ declare export var StakeCredKind: {|
   +Script: 1, // 1
 |};
 
+/**
+ */
+
+declare export var AddressHeader: {|
+  +BasePaymentKeyStakeKey: 0, // 0
```
SebastienGllmt, line 355: The values generated by Flow here are actually wrong (Typescript is correct)

Just another reason to drop Flow :/

rooooooooob: with wasm_bindgen/flow you should never be using the actual integer value of an enum. it doesn't matter what you do it always sets them to 0, 1, ... etc. I had assumed this was just a wasm_bindgen thing and that typescript was affected as well (since flow is generated from that) but that's good to know that typescript retains the explicit enum values.

rooooooooob (approved):
LGTM now

rust/src/address.rs
```diff
@@ -302,44 +302,95 @@ impl JsonSchema for Address {
     fn is_referenceable() -> bool { String::is_referenceable() }
 }
 
+#[wasm_bindgen]
+#[derive(Clone, Debug, Eq, Ord, PartialEq, PartialOrd)]
```
rooooooooob, line 306: It might be good to leave a comment here though that you shouldn't == against the header and should instead & or else the network ID could get in the way.

SebastienGllmt: Yeah you're right, this felt like a footgun so I changed the way this works a bit

rooooooooob (approved):
I think that's good now with that change/comment, as long as we drop flow support before we release 1.0.0 (so they don't think that 0, 1, 2.. etc mean anything)

PR #45, opened 2022-04-08
TxBuilder rewrite: certificates (partial)

SebastienGllmt:
The current tx builder doesn't support Plutus at all and barely supports native scripts (only through a hacky solution)

We need to annotate every step of the tx builder with information about scripts to properly build the transaction and estimate the fee. We could do this in the same style as we've done historically (just add a bunch of `set_` functions), but this is already near-unusable levels of uglyness so I'm thinking of rewriting all the set functions as builders so that we can isolate complexity of individual parts of a transaction (such as certificates) into their own builds and the main builder is just there to aggregate the data

In this PR, I wrote the CertificateBuilder, but I didn't connect it to the main tx builder. This is because if I did that, I would be stuck refactoring a huge amount of code and this PR is already big enough. Probably the best path forward is to rewrite all the chunks into these kinds of builders, then do a PR that connects them all together.

### DIFF
--- a/rust/src/builders/certificate_builder.rs
+++ b/rust/src/builders/certificate_builder.rs
@@ -39,6 +39,8 @@ pub fn add_cert_vkeys(cert_enum: &Certificate, vkeys: &mut HashSet<Ed25519KeyHas
     Ok(())
 }
 
+/// Check if the script hash matches the one in the certificate
+/// If the certificate does not require the script to sign the transaction, Ok(false) is returned
 fn check_cert_script_hash(cert_enum: &Certificate, expected_hash: &ScriptHash) -> Result<bool, JsError> {
     match &cert_enum.0 {
         // stake key registrations do not require a witness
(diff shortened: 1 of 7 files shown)


rust/src/builders/certificate_builder.rs
```diff
@@ -0,0 +1,122 @@
...
+        add_cert_vkeys(&self.cert, &mut vkey_set)?;
+        Ok(CertificateBuilderResult {
+            cert: self.cert.clone(),
+            aggregate_witness: InputAggregateWitnessData::Vkeys(vkey_set.clone()),
+        })
+    }
+
+    pub fn native_script(&self, native_script: &NativeScript) -> Result<CertificateBuilderResult, JsError> {
+        let expected_hash = native_script.hash(ScriptHashNamespace::NativeScript);
+        check_cert_script_hash(&self.cert, &expected_hash)?;
```
rooooooooob, line 106: Will we need the `Ok(bool)` return type in other spots?

rooooooooob (commented):
>so I'm thinking of rewriting all the set functions as builders so that we can isolate complexity of individual parts of a transaction (such as certificates) into their own builds and the main builder is just there to aggregate the data

I agree. The API has been getting more and more difficult to understand and that would help a lot. As would breaking it into phases when necessary like that witness builder that you started in cardano-serialization-lib.

rust/src/builders/certificate_builder.rs
SebastienGllmt: Ah, I forgot to used the variable. My bad

PR #47, opened 2022-04-19
Add more builder types

SebastienGllmt:
This adds all the component builders I think we will need

The next step after this will be changing the tx builder to have functions like `add_cert` that takes in a builder result instead of a single "add_certs" function. The reason for this change is that having a single add_certs function that deals with all the witness structure of native scripts & plutus would be really ugly, and so things are instead isolated into their own builders.

These builders support inlining witnesses as you build them up, or allow just doing `skip_witness` if, for example, you're building wallet software where the signature comes later. Native scripts & Plutus allow partially filling in the witness as well while keeping track of the # of witnesses requires so we can calculate the tx fee accurately

### DIFF
--- a/rust/src/builders/certificate_builder.rs
+++ b/rust/src/builders/certificate_builder.rs
@@ -2,6 +2,48 @@ use crate::*;
 use crate::builders::witness_builder::{InputAggregateWitnessData, PartialPlutusWitness};
 use std::collections::{HashSet};
 
+use super::witness_builder::RequiredWitnessSet;
+
+// comes from witsVKeyNeeded in the Ledger spec
+pub fn cert_required_wits(cert_enum: &Certificate, required_witnesses: &mut RequiredWitnessSet) -> () {
+    match &cert_enum.0 {
+        // stake key registrations do not require a witness
+        CertificateEnum::StakeRegistration(_cert) => (),
+        CertificateEnum::StakeDeregistration(cert) => match cert.stake_credential().kind() {
+            StakeCredKind::Script => {
+                required_witnesses.add_script_hash(&cert.stake_credential().to_scripthash().unwrap());
+            }
+            StakeCredKind::Key => {
+                required_witnesses.add_vkey_key_hash(&cert.stake_credential().to_keyhash().unwrap());
+            }
+        },
+        CertificateEnum::StakeDelegation(cert) => match cert.stake_credential().kind() {
+            StakeCredKind::Script => {
+                required_witnesses.add_script_hash(&cert.stake_credential().to_scripthash().unwrap());
+            }
+            StakeCredKind::Key => {
+                required_witnesses.add_vkey_key_hash(&cert.stake_credential().to_keyhash().unwrap());
+            }
+        },
+        CertificateEnum::PoolRegistration(cert) => {
+            for owner in &cert.pool_params().pool_owners().0 {
+                required_witnesses.add_vkey_key_hash(&owner.clone());
+            }
+            required_witnesses.add_vkey_key_hash(&cert.pool_params().operator());
+        },
+        CertificateEnum::PoolRetirement(cert) => {
+            required_witnesses.add_vkey_key_hash(&cert.pool_keyhash());
+        },
+        CertificateEnum::GenesisKeyDelegation(cert) => {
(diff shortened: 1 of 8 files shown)


rust/src/builders/input_builder.rs
```diff
@@ -0,0 +1,161 @@
...
+            required_wits: required_wits.clone(),
+        })
+    }
+
+    pub fn vkey(&self, vkey: &Vkey) -> Result<InputBuilderResult, JsError> {
+        let mut required_wits = RequiredWitnessSet::default();
+        input_required_wits(&self.utxo_info,&mut required_wits);
+        let mut required_wits_left = required_wits.clone();
+
+        let keyhash = &vkey.public_key().hash();
```
rooooooooob, line 64: it's accessed everywhere as &keyhash so we should either access it as `keyhash` directly (as it's a ref - rust magically saves temps that you take a ref of like that from going out of scope) or make it not a ref. I think this is a clippy warning too? Or one of the rust static checkers.

rust/src/builders/certificate_builder.rs
```diff
@@ -2,6 +2,48 @@ use crate::*;
 use crate::builders::witness_builder::{InputAggregateWitnessData, PartialPlutusWitness};
 use std::collections::{HashSet};
 
+use super::witness_builder::{RequiredWitnessSet, NativeScriptWitnessInfo, PlutusScriptWitnessInfo};
+
+// comes from witsVKeyNeeded in the Ledger spec
+pub fn cert_required_wits(cert_enum: &Certificate, required_witnesses: &mut RequiredWitnessSet) -> () {
```
rooooooooob, line 8: same comment about `-> ()`

rust/src/builders/certificate_builder.rs
```diff
@@ -94,41 +103,89 @@ impl SingleCertificateBuilder {
...
+        }
+    }
+
+    pub fn vkeys(&self, vkeys: &Vkeys) -> Result<CertificateBuilderResult, JsError> {
+        let mut required_wits = RequiredWitnessSet::default();
+        cert_required_wits(&self.cert, &mut required_wits);
+        let mut required_wits_left = required_wits.clone();
+
+        // the user may have provided more witnesses than required. Strip it down to just the required wits
+        let provided_wit_subset: Vec<&Vkey> = vkeys.0.iter().filter(|vkey| required_wits_left.vkeys.contains(&vkey.public_key().hash())).collect();
```
rooooooooob, line 123: same comment here about hashing

rust/src/builders/certificate_builder.rs
```diff
@@ -94,41 +103,89 @@ impl SingleCertificateBuilder {
...
+        required_wits_left.scripts.remove(&native_script.hash(ScriptHashNamespace::NativeScript));
+
+        if required_wits_left.len() > 0 {
+            return Err(JsError::from_str(&format!("Missing the following witnesses for the certificate: \n{:#?}", required_wits_left.to_str()))); 
         }
+
+        Ok(CertificateBuilderResult {
+            cert: self.cert.clone(),
+            aggregate_witness: if contains { Some(InputAggregateWitnessData::NativeScript(native_script.clone(), witness_info.clone())) } else { None },
+            required_wits: required_wits.clone(),
```
rooooooooob, line 162: same comment about cloning (I think this is in all of them)

rooooooooob (approved):
LGTM besides some minor nitpicks

rust/src/builders/withdrawal_builder.rs
```diff
@@ -0,0 +1,131 @@
...
+            required_wits: required_wits.clone(),
+        }
+    }
+
+    pub fn vkeys(&self, vkeys: &Vkeys) -> Result<WithdrawalBuilderResult, JsError> {
+        let mut required_wits = RequiredWitnessSet::default();
+        withdrawal_required_wits(&self.address, &mut required_wits);
+        let mut required_wits_left = required_wits.clone();
+
+        // the user may have provided more witnesses than required. Strip it down to just the required wits
```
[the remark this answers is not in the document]
SebastienGllmt: Good point. Not sure if there is anything that can be done about this though since the point or this PR is to move away from dummy witnesses since it doesn't work for non-Plutus input types

PR #48, opened 2022-04-20
Fix for long addresses

rooooooooob:
Ignores trailing data corresponding to input-output-hk/cardano-ledger#2729

Address saves the trailing data for consistent round-tripping on such
address(es).

Note: this PR should **never** be merged. It's just a temporary workaround until Vasil

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -406,9 +406,6 @@ impl Address {
 
     fn from_bytes_impl(data: &[u8]) -> Result<Address, DeserializeError> {
         use std::convert::TryInto;
-        const TRAILING_WHITELIST: [&[u8]; 1] = [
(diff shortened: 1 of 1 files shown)


gostkin:
@rooooooooob should we close this one?

gostkin:
> @rooooooooob should we close this one?

@rooooooooob 🙃

rooooooooob:
@gostkin Yeah I think we can close this to clean up the open PRs. This is on the old codebase only. I'll need to check the current codebase on the entire chain once we run that (very soon) to see if we need to expand the whitelist(the linked issue says there are at least 4). I believe we had this PR so we could have a branch that would never fail (would accept any trailing address data regardless of if it's in the whitelist). Just don't delete the branch for now in case I guess.

IIRC the idea was once Vasil hits that trailing data after addresses won't be allowed and we can look up in a whitelist to see if we're one of the exception addresses. @SebastienGllmt is this right?

PR #49, opened 2022-04-20
Long addr fix whitelist

rooooooooob:
Ignores trailing data corresponding to https://github.com/input-output-hk/cardano-ledger/issues/2729

Address saves the trailing data for consistent round-tripping on such
address(es).

Also adds a whitelist where only whitelisted addresses have the trailing data ignored.


SebastienGllmt (approved):
LGTM

PR #50, opened 2022-04-24
Add a test on real data

siegfried:
As we have discussed [here](https://discord.com/channels/881538386857451560/942104729725567007/967404977347772436), `add_inputs_from` may spend duplicate UTxOs when 1 UTxO is not enough to cover all the outputs. I'm trying to fix this issue in this PR.
I'll add unit test to show the issue first.

#### Fee Settings (on Testnet)
A: 44
B: 155381

#### UTxOs
1. 1 tADA
2. 10 tADA

#### Outputs
1 tADA

### DIFF
--- a/rust/src/builders/tx_builder.rs
+++ b/rust/src/builders/tx_builder.rs
@@ -3160,6 +3160,8 @@ mod tests {
             &mut output_total,
             |value| Some(value.coin),
             &mut rng).unwrap();
+        assert!(!available_indices.contains(&0));
+        assert!(available_indices.contains(&1));
         assert!(available_indices.len() < 2);
     }
 
--- a/rust/src/builders/tx_builder.rs
+++ b/rust/src/builders/tx_builder.rs
@@ -3087,6 +3087,47 @@ mod tests {
     }
 
     #[test]
+    fn tx_builder_cip2_random_improve_on_real_data() {
+        let mut tx_builder = create_tx_builder_with_fee(&create_linear_fee(44, 155381));
+        const COST: u64 = 1000000;
+        tx_builder.add_output(
+            &TransactionOutputBuilder::new()
+                .with_address(&Address::from_bech32("addr1vyy6nhfyks7wdu3dudslys37v252w2nwhv0fw2nfawemmnqs6l44z").unwrap())
+                .next().unwrap()
+                .with_coin(&to_bignum(COST))
+                .build().unwrap()
+            ).unwrap();
+        let mut available_inputs = TransactionUnspentOutputs::new();
+        available_inputs.add(&make_input(0u8, Value::new(&to_bignum(1000000))));
+        available_inputs.add(&make_input(1u8, Value::new(&to_bignum(10000000))));
+        let add_inputs_res =
+            tx_builder.add_inputs_from(&available_inputs, CoinSelectionStrategyCIP2::RandomImprove);
+        assert!(add_inputs_res.is_ok(), "{:?}", add_inputs_res.err());
+        let change_addr = ByronAddress::from_base58("Ae2tdPwUPEZGUEsuMAhvDcy94LKsZxDjCbgaiBBMgYpR8sKf96xJmit7Eho").unwrap().to_address();
+        let add_change_res = tx_builder.add_change_if_needed(&change_addr);
+        assert!(add_change_res.is_ok(), "{:?}", add_change_res.err());
+        let tx_build_res = tx_builder.build();
+        assert!(tx_build_res.is_ok(), "{:?}", tx_build_res.err());
+        let tx = tx_build_res.unwrap();
+        // we need to look up the values to ensure there's enough
+        let mut input_values = BTreeMap::new();
+        for utxo in available_inputs.0.iter() {
+            input_values.insert(utxo.input.transaction_id(), utxo.output.amount.clone());
+        }
+        let mut encountered = std::collections::HashSet::new();
+        let mut input_total = Value::new(&Coin::zero());
+        for input in tx.inputs.0.iter() {
+            let txid = input.transaction_id();
+            if !encountered.insert(txid.clone()) {
+                panic!("Input {:?} duplicated", txid);
+            }
+            let value = input_values.get(&txid).unwrap();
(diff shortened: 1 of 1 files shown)


rust/src/builders/tx_builder.rs
```diff
@@ -3086,6 +3086,47 @@ mod tests {
         assert!(input_total >= Value::new(&tx_builder.min_fee().unwrap().checked_add(&to_bignum(COST)).unwrap()));
     }
 
+    #[test]
+    fn tx_builder_cip2_random_improve_on_real_data() {
```
siegfried, line 3090: Note that this test doesn't always fail due to the randomness. When it fails, it returns as follow:

```
failures:

---- builders::tx_builder::tests::tx_builder_cip2_random_improve_on_real_data stdout ----
thread 'builders::tx_builder::tests::tx_builder_cip2_random_improve_on_real_data' panicked at 'Input TransactionHash([0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]) duplicated', src/builders/tx_builder.rs:3122:17


failures:
    builders::tx_builder::tests::tx_builder_cip2_random_improve_on_real_data

test result: FAILED. 165 passed; 1 failed; 1 ignored; 0 measured; 0 filtered out; finished in 4.30s

error: test failed, to rerun pass '--lib'
```

siegfried:
The root cause is that `cip2_random_improve_by()` doesn't exclude the chosen index from `available_indices`, which could [be chosen again](https://github.com/dcSpark/cardano-multiplatform-lib/blob/366429478486c774fdb993f5d6698f3818821c5f/rust/src/builders/tx_builder.rs#L314) by the random algorithm.

rust/src/builders/tx_builder.rs
siegfried: It is a fifty-fifty chance. It fails when 10 ADA is selected first.

rust/src/builders/tx_builder.rs
```diff
@@ -485,19 +485,23 @@ impl TransactionBuilder {
...
+                        let max = 3 * min;
+                        let move_closer = (ideal as i128 - new as i128).abs() < (ideal as i128 - cur as i128).abs();
+                        let not_exceed_max = new < max;
+
+                        move_closer && not_exceed_max
+                    };
+                    if should_improve {
                         available_indices.insert(*i);
                         available_indices.remove(j);
+                        std::mem::swap(i, j);
```
siegfried, line 504: This should fix it. New input should be removed from available UTxO set whereas the old one should be put back. The order didn't make sense. @SebastienGllmt

SebastienGllmt: @rooooooooob

rust/src/builders/tx_builder.rs
SebastienGllmt: We should avoid tests that fail randomly. I would feel more comfortable if we changed the function to instead take in a seed to use for the randomness so that by default it's random, but given a seed it will provide a deterministic result. It would help users in some cases and also help make our tests deterministic.

siegfried: With the fix it's 100% passed. However, I can remove this and test with a seed in [here](https://github.com/dcSpark/cardano-multiplatform-lib/pull/50/files#diff-f9e91e0f5b08343e5a3f64238729812bc8daf2f2beabc959064832b61d0d7d05R3155).

siegfried: I created PR #51 to upgrade `rand` and `rand_chacha`. I'll make the change after that is merged. @SebastienGllmt

rust/src/builders/tx_builder.rs
```diff
@@ -426,17 +425,16 @@ impl TransactionBuilder {
...
-    fn cip2_random_improve_by<F>(
+    fn cip2_random_improve_by<F, R: Rng + ?Sized>(
         &mut self,
         available_inputs: &[TransactionUnspentOutput],
         available_indices: &mut BTreeSet<usize>,
         input_total: &mut Value,
         output_total: &mut Value,
         by: F,
-        rng: &mut rand::rngs::ThreadRng) -> Result<(), JsError>
+        rng: &mut R) -> Result<(), JsError>
```
siegfried, line 435: By changing the type to `rand::Rng` trait it is possible to use `ChaChaRng` from constant seeds in unit tests. However, current `ChaChaRng` of `rand_chacha` implements [`rand_os::rand_core::RngCore`](https://github.com/dcSpark/cardano-multiplatform-lib/blob/d9ec0ce32136deda2a497dd1c04e6dce8932ba5a/rust/src/chain_crypto/testing.rs#L6) instead of `rand::RngCore`, which cannot pass the [restriction](https://docs.rs/rand/latest/rand/trait.Rng.html) of `: rand::RngCore`. Any idea how to get around of this without upgrading? @SebastienGllmt

SebastienGllmt: yeah I remember all these crates have to be aligned at the same version to work properly by design so the upgrade is required

rust/src/builders/tx_builder.rs
rooooooooob: Good catch @siegfried

siegfried: After #53 is merged, this is all set. Or merge this directly since it has already rebased on that one.

rooooooooob: I think that I started writing it another way then later changed to indices halfway through since it made something easier. I forgot what though.

siegfried: Ok, thanks. I think this is ready to merge. I can create another PR to refactor if it is needed.

rooooooooob (approved):
LGTM

siegfried:
@SebastienGllmt Is there any other change required? The warnings are not from this PR. I tried to remove the warnings but some them such as "warning: field is never read: `cert`" are something you are still working on?

PR #51, opened 2022-04-25
Update crates

siegfried:
* rand: 0.8.4 -> 0.8.5
* rand_chacha: 0.1.1 -> 0.3.1
* rand_os: removed in favor of rand

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -13,15 +13,6 @@ dependencies = [
 
 [[package]]
 name = "autocfg"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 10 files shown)


SebastienGllmt:
I remember the issue with the rand crates is that they don't always work with browser and nodejs builds. Not sure if maybe this upgrade impacts this (I think notably it worked on browsers but not nodejs or something like that)

siegfried:
I believe it is dedicated to `OsRng` since you are already use `rand` a lot?

siegfried:
Do you have a reference of the issue? Is there a way to reproduce it?

SebastienGllmt (approved):

PR #53, opened 2022-04-25
Fix no field error in Address

siegfried:


siegfried:
The compilation is broken. Any plan to setup test actions? @SebastienGllmt

rooooooooob (approved):
Yeah this was an issue I think from #49 and #47 conflicting. Thanks for the fix.

SebastienGllmt:
We do that test actions (in fact, they ran on this PR). I think it's just that they don't enforce rebasing before a PR is merged so the issue wasn't caught

PR #57, opened 2022-04-30
Update TX builder

siegfried:


siegfried:
Still WIP. Feel free to let me know if you find anything doesn't make sense. @SebastienGllmt

rust/src/builders/witness_builder.rs
```diff
@@ -303,6 +303,54 @@ impl TransactionWitnessSetBuilder {
...
+                match info.0 {
+                    NativeScriptWitnessInfoKind::Count(num) => self.add_vkeys_by_num(num),
+                    NativeScriptWitnessInfoKind::Vkeys(ref vkeys) => self.add_vkeys(vkeys),
+                    NativeScriptWitnessInfoKind::AssumeWorst => {
+                        let num = script.get_required_signers().len();
+                        self.add_vkeys_by_num(num);
+                    }
+                }
+            }
+            InputAggregateWitnessData::PlutusScriptNoDatum(witness, info) => {
```
siegfried, line 338: I'm not sure about the Plutus part. Am I doing right? @SebastienGllmt

rust/src/builders/tx_builder.rs
```diff
@@ -4029,33 +3567,63 @@ mod tests {
...
+            );
+            builder.native_script(&mint_script1, &NativeScriptWitnessInfo::assume_signature_count()).unwrap()
+        };
+        tx_builder.add_input(&input);
 
         // Original tx fee now assumes two VKey signatures for two inputs
         let original_tx_fee = tx_builder.min_fee().unwrap();
-        assert_eq!(original_tx_fee, to_bignum(168361));
+        // The original test did not include native script
+        // assert_eq!(original_tx_fee, to_bignum(168361));
```
siegfried, line 3596: Now all tests are passed except this one. I'm not sure if the numbers used by this test were correct since the origin test didn't include the native script in the witness set. @SebastienGllmt @rooooooooob

rust/src/builders/witness_builder.rs
```diff
@@ -303,6 +303,54 @@ impl TransactionWitnessSetBuilder {
...
+    fn add_vkeys(&mut self, vkeys: &Vec<Vkey>) {
+        let fake_sig = fake_raw_key_sig(0);
+        for vkey in vkeys {
+            let fake_vkey_witness = Vkeywitness::new(&vkey, &fake_sig);
+            self.add_vkey(&fake_vkey_witness);
+        }
+    }
+
+    fn add_vkeys_by_num(&mut self, num: usize) {
+        let vkeys: Vec<Vkey> = (0..num).into_iter().map(|_| Vkey::new(&fake_raw_key_public(0))).collect();
```
SebastienGllmt, line 315: If you're always using the same fake raw key, it will just get deduplicated so it won't properly count towards the tx fee like you want it to.
You could fix this by using the iterator index instead of hardcoding `0`, but this still won't work if `add_vkeys_by_num` gets called twice for the same transaction

Additionally, right now this has a chance of double-counting witnesses. Imagine a transaction that contains 2 Plutus scripts both which have the same required signers. This code would double the amount of fake witnesses added even though a single copy would be enough

I think the proper fix to this would be to: instead of calling this function in multiple places, instead:
1. deduplcate the ED25519 keys from missing signers
    1. easy version: deduplicate missing signers only.
    2. More complicated version I don't think we want: deduplicate missing signers and also remove any missing signer that was added somewhere else as a `vkeys`. Presumably the case this happens is low and it's kind of a user error, so only doing (1) is good enough I think
3. Call this function setting `num` as `deduplicated.len()` + the sum of all the `NativeScriptWitnessInfoKind::Count(num)`

rust/src/builders/witness_builder.rs
SebastienGllmt: Redeemers don't go in the `add_plutus_datum`. Instead, you can use `add_redeemer`. The partial redeemers need to be combined with an index and a tag. The idea is that we can know the tag and the index so the user doesn't have to pass them in (ex: if they called `add_certificate`, the index is next available position in the certificates array and the tag is "Cert". You should be able to complete these partial redeemers so that they are full redeemers from within the context  of the tx_builder

rust/src/builders/tx_builder.rs
```diff
@@ -509,89 +399,33 @@ impl TransactionBuilder {
...
-                }
-            },
-            None => (),
-        }
-        match &ByronAddress::from_address(address) {
-            Some(addr) => self.add_bootstrap_input(addr, input, amount),
-            None => (),
-        }
+    pub fn add_utxo(&mut self, result: &InputBuilderResult) {
+        self.utxos.push(result.clone());
```
SebastienGllmt, line 424: It feels kind of awkward that this and `add_input` are separate functions. Probably this awkwardness comes from the fact the input selection is too tightly coupled with the tx builder, but probably this should be separated in a follow-up PR. If you really want to include this function, maybe add a warning / reminder that we want to delete this at some point.

rust/src/builders/witness_builder.rs
```diff
@@ -338,15 +338,41 @@ impl TransactionWitnessSetBuilder {
             InputAggregateWitnessData::PlutusScriptNoDatum(witness, info) => {
                 self.add_plutus_script(&witness.script());
                 self.add_plutus_datum(&witness.untagged_redeemer().datum());
-                self.add_fake_vkey_witnesses_by_num(info.missing_signers.len());
-                self.add_fake_vkey_witnesses(&info.known_signers.0);
+
+                let mut required_wits = RequiredWitnessSet::new();
```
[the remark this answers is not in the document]
siegfried: Missing signers are deduplicated after the key hashes are given to `required_wits`. This [code](https://github.com/dcSpark/cardano-multiplatform-lib/pull/57/files#diff-0318c1c246c0c5d8f74f1b087e11b7e27f27f353c2443ecc886914dca673db44R354) should be able to resolve multiple calls of `add_input_aggregate_witness_data`.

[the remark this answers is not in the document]
siegfried: The `required_wits` is updated [here](https://github.com/dcSpark/cardano-multiplatform-lib/pull/57/files#diff-0318c1c246c0c5d8f74f1b087e11b7e27f27f353c2443ecc886914dca673db44R351).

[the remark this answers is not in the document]
siegfried: A related test is added [here](https://github.com/dcSpark/cardano-multiplatform-lib/pull/57/files#diff-0318c1c246c0c5d8f74f1b087e11b7e27f27f353c2443ecc886914dca673db44R508).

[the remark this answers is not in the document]
siegfried: > Additionally, we don't need fake witnesses for known_signers because we know what the real key is.

Still need to calculate the size for fee, right? Otherwise where is supposed to fetch the real witness?

rust/src/builders/tx_builder.rs
siegfried: Yeah, actually I'm going to move CIP2 out or implement [this](https://github.com/cardano-foundation/CIPs/issues/232) instead after this PR is merged.

rust/src/builders/witness_builder.rs
siegfried: I've used `add_redeemer` in `add_*` of transaction builder instead. Please review.

rust/src/builders/tx_builder.rs
```diff
@@ -662,25 +507,44 @@ impl TransactionBuilder {
...
-            add_cert_vkeys(cert, &mut vkey_set)?;
-        };
-        for vkey in &vkey_set {
-            self.input_types.vkeys.insert(vkey.clone());
+    pub fn add_cert(&mut self, result: &CertificateBuilderResult) {
+        let mut certs = self.certs.clone().unwrap_or(Certificates::new());
+        certs.add(&result.cert);
+        self.certs = Some(certs);
+        if let Some(ref data) = result.aggregate_witness {
+            self.witness_set_builder.add_input_aggregate_witness_data(data);
```
SebastienGllmt, line 515: Note to self: verify these fake witnesses are only used later for fee estimation and are separate from the real witnesses we add on the build() call

rust/src/builders/tx_builder.rs
```diff
@@ -740,121 +599,35 @@ impl TransactionBuilder {
...
-                witness_scripts.add(&policy_script.clone());
+    pub fn add_mint(&mut self, result: &MintBuilderResult) {
+        if let Some(ref data) = result.aggregate_witness {
+            self.witness_set_builder.add_input_aggregate_witness_data(data);
+            if let InputAggregateWitnessData::PlutusScript(witness, _, _) = data {
+                let redeemer = {
+                    let untagged = witness.untagged_redeemer();
+                    let data = untagged.datum();
+                    let ex_units = untagged.ex_units();
+                    let index = self.mint.as_ref().and(Some(1u64)).or(Some(0)).unwrap();
```
SebastienGllmt, line 610: Are you sure about this? If yes, probably we should add a comment explaining this.

rust/src/builders/tx_builder.rs
```diff
@@ -867,18 +640,14 @@ impl TransactionBuilder {
...
-                vkeys: BTreeSet::new(),
-                scripts: BTreeSet::new(),
-                bootstraps: BTreeSet::new(),
-            },
             validity_start_interval: None,
             mint: None,
-            native_scripts: None,
             script_data_hash: None,
             collateral: None,
             required_signers: None,
```
SebastienGllmt, line 647: We have to redesign `required_signers` because we can't just call `set_required_signers` anymore. It has to take into account the fact that some required signers may already have been added from the input/mint/cert/withdrawal builders

rust/src/builders/tx_builder.rs
```diff
@@ -1276,25 +1045,13 @@ impl TransactionBuilder {
...
-    }
-
     /// Returns full Transaction object with the body and the auxiliary data
     /// NOTE: witness_set will contain all mint_scripts if any been added or set
     /// NOTE: is_valid set to true
     pub fn build_tx(&self) -> Result<Transaction, JsError> {
         Ok(Transaction {
             body: self.build()?,
-            witness_set: self.get_witness_set(),
+            witness_set: self.witness_set_builder.build()?,
```
SebastienGllmt, line 1054: This won't work properly because this will also include all the useless fake witnesses we created purely for tx fee calculation

rust/src/builders/witness_builder.rs
```diff
@@ -303,6 +302,58 @@ impl TransactionWitnessSetBuilder {
...
+            self.add_vkey(&fake_vkey_witness);
+        }
+    }
+
+    fn add_fake_vkey_witnesses_by_num(&mut self, num: usize) {
+        let vkeys: Vec<Vkey> = (0..num).into_iter().map(|i| Vkey::new(&fake_raw_key_public((i + self.vkeys.len()) as u8))).collect();
+        self.add_fake_vkey_witnesses(&vkeys);
+    }
+
+    pub fn add_input_aggregate_witness_data(&mut self, data: &InputAggregateWitnessData) {
```
SebastienGllmt, line 318: This function won't work in WASM because this struct (`TransactionWitnessSetBuilder`) is visible for wasm, but this member function has an argument `InputAggregateWitnessData` which isn't visible from WASM (and note that this enum can't easily be exposed to WASM either)

Since this function never needs to be called from WASM, you may want to just make it so it's not a member function (just place it somewhere else in the file)

The proper way to fix this would have been to make that all of `TransactionWitnessSetBuilder` is a non-WASM Rust-only struct (and therefore supports this function) and then expose a WASM-compatible subset of the struct

siegfried: How about making the function public only in this crate, like those attributes in other builders. This function shouldn't be visible anyway.

rust/src/builders/tx_builder.rs
siegfried: No, I'm not. On second thought this index should be the length of assets.

rust/src/builders/witness_builder.rs
siegfried: Updated.

rust/src/builders/tx_builder.rs
siegfried: Updated.

rust/src/builders/witness_builder.rs
```diff
@@ -303,6 +302,58 @@ impl TransactionWitnessSetBuilder {
...
+    fn add_fake_vkey_witnesses(&mut self, vkeys: &Vec<Vkey>) {
+        let fake_sig = fake_raw_key_sig(0);
+        for vkey in vkeys {
+            let fake_vkey_witness = Vkeywitness::new(&vkey, &fake_sig);
+            self.add_vkey(&fake_vkey_witness);
+        }
+    }
+
+    fn add_fake_vkey_witnesses_by_num(&mut self, num: usize) {
+        let vkeys: Vec<Vkey> = (0..num).into_iter().map(|i| Vkey::new(&fake_raw_key_public((i + self.vkeys.len()) as u8))).collect();
```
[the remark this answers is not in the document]
siegfried: Added.

rust/src/builders/witness_builder.rs
```diff
@@ -303,6 +302,58 @@ impl TransactionWitnessSetBuilder {
...
+        }
+    }
+
+    fn add_plutus_witness_info(&mut self, info: &PlutusScriptWitnessInfo) {
+        let known_signers = &info.known_signers.0;
+        let missing_signers = &info.missing_signers.0;
+
+        self.add_fake_vkey_witnesses(known_signers);
+        self.add_fake_vkey_witnesses_by_num(
+            missing_signers.iter().filter(|hash| !self.required_wits.vkeys.contains(&hash)).count()
```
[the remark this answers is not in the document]
siegfried: Added.

rust/src/builders/tx_builder.rs
siegfried: So I'll just remove it.

rust/src/builders/tx_builder.rs
siegfried: In fact, this will include no fake witness because it would fail at remaining witnesses check. An error is returned instead. However, this prevents the fee being calculated correctly. I've extract the `build()` for fee calculation. Please review.

rust/src/builders/tx_builder.rs
siegfried: BTW, how is `required_signers` used in transaction body? We've never set it in our app and there is no unit tests against it. My current implement is to convert remaining required witnesses into `required_signers`. Does it make sense?

rust/src/builders/witness_builder.rs
```diff
@@ -360,43 +360,55 @@ impl TransactionWitnessSetBuilder {
         );
     }
 
-    pub fn build(&self) -> Result<TransactionWitnessSet, JsError> {
+    pub(crate) fn build(&self) -> TransactionWitnessSet {
```
SebastienGllmt, line 377: I don't think we can make this function `pub(crate)` because we need this to be visible to Rust projects that import CML

rust/src/builders/tx_builder.rs
```diff
@@ -600,29 +600,31 @@ impl TransactionBuilder {
...
+        self.mint = Some(mint);
         if let Some(ref data) = result.aggregate_witness {
             self.witness_set_builder.add_input_aggregate_witness_data(data);
             if let InputAggregateWitnessData::PlutusScript(witness, _, _) = data {
                 let redeemer = {
                     let untagged = witness.untagged_redeemer();
                     let data = untagged.datum();
                     let ex_units = untagged.ex_units();
-                    let index = self.mint.as_ref().and(Some(1u64)).or(Some(0)).unwrap();
+                    let index = self.mint.as_ref()
```
SebastienGllmt, line 618: Note to self: look into whether or not this is correct since there is still no comment here explaining whether or not this is really how mint indices are supposed to work

rust/src/builders/witness_builder.rs
```diff
@@ -388,7 +402,7 @@ impl TransactionWitnessSetBuilder {
         result
     }
 
-    fn remaining_wits(&self) -> RequiredWitnessSet {
+    pub(crate) fn remaining_wits(&self) -> RequiredWitnessSet {
```
SebastienGllmt, line 405: Again, let's not make functions `pub(crate)` unless we have a reason to

rust/src/builders/witness_builder.rs
```diff
@@ -113,6 +113,20 @@ pub struct RequiredWitnessSet {
     pub(crate) redeemers: HashSet<RedeemerWitnessKey>,
 }
 
+impl From<RequiredWitnessSet> for Option<RequiredSigners> {
+    fn from(wits: RequiredWitnessSet) -> Self {
+        let mut required_signers = RequiredSigners::new();
+
+        wits.vkeys.union(&wits.bootstraps).for_each(|hash| required_signers.add(hash));
```
SebastienGllmt, line 120: Remaining signers is a special field inside transactions. It's not just the sum of whatever witness needs to be added. Notably, it's the combination of all the `missing_signers` and `known_signers`

rust/src/builders/tx_builder.rs
siegfried: Minting indices are the total length of minting assets. It cannot work like others when minting results are added in a batch. But the indices should be identical.

siegfried: Or wrap Mint in vector in transaction builder?

rust/src/builders/witness_builder.rs
siegfried: When `NativeScriptWitnessInfoKind::Count(num)` is used, the public keys are fake, so the key hashes should not be assign to `required_signers`. How about keep `TransactionBuilder.set_required_signers()` and let the users set it explicitly? It is not mandatory in real use anyway.

siegfried: Rolled back for now.

rust/src/builders/witness_builder.rs
siegfried: updated

rust/src/builders/witness_builder.rs
siegfried: Updated.

rust/src/builders/tx_builder.rs
siegfried: Ok, this is incorrect. I think I've figure it out. Will change it soon.

siegfried: Fixed and added some unit tests.

rust/src/builders/witness_builder.rs
```diff
@@ -277,86 +286,159 @@ impl TransactionWitnessSetBuilder {
...
-        };
-        match &wit_set.plutus_data() {
-            None => (),
-            Some(plutus_data) => plutus_data.elems.iter().for_each(|plutus_datum| { self.add_plutus_datum(plutus_datum); } ),
-        };
-        match &wit_set.redeemers() {
-            None => (),
-            Some(redeemers) => redeemers.0.iter().for_each(|redeemer| { self.add_redeemer(redeemer); } ),
-        };
+        if let Some(vkeys) = wit_set.vkeys() {
```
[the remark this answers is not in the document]
rooooooooob: Yeah it'd make more sense performance-wise to use `vkeys` directly and avoid the clone since we're using it from within the same crate. The clone on `vkeys()` is due to wasm since wasm_bindgen gets really ugly/confusing for users if you don't clone or if you take inputs by-value instead of by-ref since there is no rustc on the JS side telling you when something has been moved out or not, leading to confusing crashes if you're not careful or not aware of wasm_bindgen details (and I imagine most js users are not).

We end up cloning it later on in the function anyway with `add_vkey` so you're right there. In the current code it would be cloned twice.

In general performance isn't the hugest impact here since it's utterly and completely dwarfed by any cryptographic operations. If you profile it you'll see just a couple hash/sign functions taking up 95%+ of the entire process of using the tx builder. Avoiding recomputing hashes/etc is a much, much more important optimization for us.

I'm working on splitting up the library into a native rust version with separate wasm bindings so that's another reason why I wouldn't worry about performance issues that are using the wasm API in the tx builder right now. At first we might just only expose this in WASM but later on we'll migrate this code to be purely in rust just with a separate thin wasm_bindgen wrapper over top of it, so issues like this will get taken care of when we do that work.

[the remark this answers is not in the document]
siegfried: Extracting a pure rust crate is a good idea. Things like coin selection can be shared with Ergo or other eUTXO chain I presume? Do you need any help on that? @rooooooooob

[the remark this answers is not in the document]
rooooooooob: @siegfried we might need help later on after the main split. The phase I'm working on right now is updating the [`cddl-codegen`](https://github.com/Emurgo/cddl-codegen) tool which I initially used to generate most of the content inside of `serialization.rs`/`lib.rs`/`plutus.rs` in the first place. After re-generating the core structs via the new [`babbage.cddl`](https://github.com/input-output-hk/cardano-ledger/blob/master/eras/babbage/test-suite/cddl-files/babbage.cddl) binary spec for the new hardfork we should also support preserving abitrary CBOR encodings (since e.g. a CBOR length/int 5 could be inlined into the type or in 1/2/4/8 additional bytes, there are 2 ways to encode lengths (definite/indefinite followed by a break code), maps can vary in order, etc) as well as the split into a rust crate and a wasm wrapper crate instead of trying to support both at once. This should make the rust user experience a lot less awful.

After I do this main split via `cddl-codegen` we'll still need to adapt a lot of the handwritten code (in particular `tx_builder.rs`, `utils.rs`, etc but also some of the hand-edits to `lib.rs`) afterwards which is where there might be some work to be done, especially in updating the tx builder to be just in rust. We'll have to see how this comes out. I hope to start the regeneration process sometime next week. In most cases it's just passing the spec through the codegen but there are some things that either a) aren't well supported or b) generate awkward APIs so those I'll need to do by hand.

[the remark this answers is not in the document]
siegfried: > since wasm_bindgen gets really ugly/confusing for users if you don't clone or if you take inputs by-value instead of by-ref since there is no rustc on the JS side telling you when something has been moved out or not, leading to confusing crashes if you're not careful or not aware of wasm_bindgen details (and I imagine most js users are not).

Is it because the argument given to the function is a reference?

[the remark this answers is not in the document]
siegfried: I presume the functions exposed as WASM should use the object instead of reference?

siegfried:
I added some low level setters and getters in case the high level `add_*` does not work.

siegfried:
Any update needed? @SebastienGllmt @rooooooooob

rust/src/builders/witness_builder.rs
```diff
@@ -263,6 +262,16 @@ impl TransactionWitnessSetBuilder {
         );
     }
 
+    pub fn add_untagged_redeemer(&mut self, tag: &RedeemerTag, untagged: &UntaggedRedeemer) {
```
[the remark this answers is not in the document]
siegfried: `indexOf` in section 4.1 needs the sets sorted, so it might not be possible to decide the indices before building the transaction.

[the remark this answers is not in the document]
siegfried: Maybe we could use a `RedeemerBuilder`.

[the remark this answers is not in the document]
siegfried: Added a `RedeemerSetBuilder` in `rust/src/builders/witness_builder.rs` and updated the `add_*` methods in transaction builder.

SebastienGllmt (approved):
LGTM. Just need to find the right time to merge this.

Also, I'm considering maybe having this new interface side-by-side with the old tx builder to make it easier for projects to move to CML during Vasil migration

siegfried:
The fee model of Plutus needs to be added after this PR is merged as well. This PR has been too large to work on.

mathieu2em:
Hello, reaching out to ask a noob question, please dont judge me for this, but why did you remove the native scripts ? Seems like the Lucid implementation is still using it to build its transactions ? Are the native scripts not supported anymore ? Is there a better way to do it now ? Thank you so much for your time and help !
@SebastienGllmt

SebastienGllmt:
@mathieu2em not sure what you mean. Native scripts are still supported

siegfried:
@mathieu2em what made you think so and why did you comment this PR?

mathieu2em:
I want to thank you for your swift response to my previous inquiries. I must apologize for the confusion in my previous message. I was exploring the use of TransactionBuilder, which is derived from the Typescript @dcspark/cardano-multiplatform-lib-browser library. Specifically, I came across the ".add_native_script()" method while analyzing how Lucid creates their transactions. However, upon instantiating the TransactionBuilder object from the multiplatform library, I could not locate the aforementioned method. After some investigation, I discovered that it had been removed from the TransactionBuilder in a recent commit.

On how I stumbled upon this PR, I was searching native script discussions in the issues of the repository. Apologies if I should have done otherwise.

I truly appreciate your assistance in shedding light on this matter. I hope that I am not causing any inconvenience and am grateful for your willingness to help.

Thank you once again for your time and support.
@SebastienGllmt @siegfried

mathieu2em:
> @mathieu2em not sure what you mean. Native scripts are still supported

Oh yeah my message was absolutely unclear, I talk about the removal of the add_native_script() method applied to Transaction Builder in this commit : 32e1865b976d51069dea82db3d9796a56783de55 
I'm pretty sure you might have replaced it by other methods that make more sense since the changes of the last months on Cardano.

siegfried:
@mathieu2em yeah, this PR changed the interfaces of transaction builder. Regarding Lucid I think they are forking CML so there are some gaps IIRC. You can check the usage in [our project](https://github.com/ADAOcommunity/round-table/blob/1a0794a0ad4757ac88f25936d13c21135bb93624/src/components/transaction.tsx#L1146) or the builders in this project.

mathieu2em:
> 
@siegfried 
I will definitely check the usage in your project, which I didnt know about and is definitely extremely interesting!

Also if I may ask one last question: Why is there a browser and a Node version of the Multiplatform-lib package ? 

Thank you very much for your answer!

siegfried:
Because Rust compile WASM for browser and Node differently. @mathieu2em

mathieu2em:
> Because Rust compile WASM for browser and Node differently. @mathieu2em

Thank you very much!

PR #58, opened 2022-05-01
Add Byron genesis parsing

SebastienGllmt:
This copies the genesis parsing code from cardano-rust and puts in it CML so that we can parse genesis blocks in projects like oura-postgres-sink. We had to do this because db-sync didn't expose all the information about the genesis block that we needed

Note: I didn't add WASM bindings to any of this since I'm not sure if this is something we'll need from WASM. Things can be exported later if needed


rust/json-gen/src/main.rs
```diff
@@ -109,7 +109,8 @@ fn main() {
     gen_json_schema!(AuxiliaryDataHash);
     gen_json_schema!(PoolMetadataHash);
     gen_json_schema!(VRFKeyHash);
-    gen_json_schema!(BlockHash);
```
SebastienGllmt, line 112: Just FYI, this was actually a type mistake in the old bindings. Block body hash and block header hash are different things (but coincidentally both happen to be 32 bytes so it didn't cause any issue)

rust/src/address.rs
```diff
@@ -36,6 +36,7 @@ fn variable_nat_encode(mut num: u64) -> Vec<u8> {
 #[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
 pub struct NetworkInfo {
     network_id: u8,
+    // FIXME: should by i32?
```
SebastienGllmt, line 39: This comment is wrong. I took this from the cardano-rust codebase, but in the CDDL for both Byron & Shelley it's an u32

rust/src/chain_crypto/hash.rs
```diff
@@ -135,9 +137,96 @@ macro_rules! define_hash_object {
                 bech32::to_bech32_from_bytes::<Self>(self.as_ref())
             }
         }
+        impl cbor_event::de::Deserialize for $hash_ty {
```
SebastienGllmt, line 140: No idea about how to add serializable types. Please review @rooooooooob

rust/src/chain_crypto/key.rs
```diff
@@ -254,6 +270,21 @@ impl<A: AsymmetricKey> Bech32 for SecretKey<A> {
...
+    }
+}
+
+impl<A: AsymmetricPublicKey> cbor_event::de::Deserialize for PublicKey<A> {
+    fn deserialize<R: std::io::BufRead>(raw: &mut Deserializer<R>) -> cbor_event::Result<Self> {
+        let result = PublicKey::<A>::from_binary(raw.bytes()?.as_ref())
+            .map_err(|err| cbor_event::Error::CustomError(format!("{}", err)))?;
+        Ok(result)
+    }
+}
```
SebastienGllmt, line 285: No idea about how to add serializable types. Please review @rooooooooob

rust/src/chain_crypto/sign.rs
```diff
@@ -244,3 +254,32 @@ pub(crate) mod test {
...
+        serializer.serialize_str(&hex::encode(&self.as_ref()))
+    }
+}
+
+impl<'de, U, A: VerificationAlgorithm> serde::de::Deserialize<'de> for Signature<U, A> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where
+    D: serde::de::Deserializer<'de> {
+        let s = <String as serde::de::Deserialize>::deserialize(deserializer)?;
+        Signature::<U, A>::from_str(&s).map_err(|_e| serde::de::Error::invalid_value(serde::de::Unexpected::Str(&s), &"hex bytes for signature"))
+    }
```
SebastienGllmt, line 284: No idea about how to add serializable types. Please review @rooooooooob

rust/src/genesis/byron/config.rs
```diff
@@ -0,0 +1,151 @@
...
+/// ```
+/// use cardano_multiplatform_lib::genesis::byron::config::ProtocolMagic;
+///
+/// assert_eq!(ProtocolMagic::default(), ProtocolMagic::new(0x2D964A09));
+/// ```
+///
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[cfg_attr(feature = "generic-serialization", derive(Serialize, Deserialize))]
+#[repr(C)]
+pub struct ProtocolMagic(u32); // FIXME: should be i32
```
SebastienGllmt, line 36: Again, u32 is correct so I don't know why this comment existed in cardano-rust

rust/src/genesis/byron/config.rs
```diff
@@ -0,0 +1,151 @@
...
+        let v = reader.unsigned_integer()? as u32;
+        Ok(ProtocolMagic::from(v))
+    }
+}
+
+#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Clone, Copy)]
+#[cfg_attr(feature = "generic-serialization", derive(Serialize, Deserialize))]
+pub enum NetworkMagic {
+    NoMagic,
+    Magic(u32), // FIXME: should by i32
```
SebastienGllmt, line 83: Again, incorrect comment

rust/src/genesis/byron/config.rs
```diff
@@ -0,0 +1,151 @@
...
+        NetworkMagic::from(*pm)
+    }
+}
+
+impl From<u32> for NetworkMagic {
+    fn from(pm: u32) -> Self {
+        // FIXME: is there a better way to determine whether to emit
+        // NetworkMagic? There is a requiresNetworkMagic field in
+        // lib/configuration.yaml, but not in the genesis data.
+        if pm == *ProtocolMagic::default() || pm == 633343913 {
```
SebastienGllmt, line 97: I think this `633343913` check can be removed. This is because both mainnet & staging didn't contain the protocol magic, but staging no longer exists (was replaced by testnet which does use it)

rust/src/genesis/byron/config.rs
```diff
@@ -0,0 +1,126 @@
...
+    // FIXME: genesis_prev shouldn't be here since it's computed *from* the GenesisData.
+    pub genesis_prev: BlockHeaderHash,
+    pub epoch_stability_depth: usize, // a.k.a. 'k'
+    pub start_time: SystemTime,
+    pub slot_duration: Duration,
+    pub protocol_magic: ProtocolMagic,
+    pub fee_policy: fees::LinearFee,
+    pub avvm_distr: BTreeMap<chain_crypto::PublicKey<Ed25519>, utils::Coin>, // AVVM = Ada Voucher Vending Machine
+    // TODO: convert to ByronAddress without affecting order
+    pub non_avvm_balances: BTreeMap<legacy_address::Addr, utils::Coin>,
```
SebastienGllmt, line 114: Note: this gives a different order than what db-sync gives. Not sure which one is correct

SebastienGllmt: https://github.com/input-output-hk/cardano-db-sync/issues/1107

rust/src/chain_crypto/hash.rs
rooooooooob: We actually use the trait in [`utils::Deserialize`](https://github.com/dcSpark/cardano-multiplatform-lib/blob/develop/rust/src/utils.rs#L963) trait to contain additional error information and such, but that trait is automatically implemented by anything that implements `cbor_event::de::Deserialize` so this is fine. It will wrap it in `DeserializeFailure::CBOR` and will contain `None` for location information (i.e. won't tell you in what part of the CBOR it was found) For simple types like a hash this isn't an issue since there's no member name to tag it with anyway, so you're only losing the struct name being tagged but honestly I don't think is a big deal.

For serialization we directly use the `cbor_event` trait so this is exactly what you need to do.

Keep in mind that if you want any of these types to **directly** have [`to_bytes()`/`from_bytes()`](https://github.com/dcSpark/cardano-multiplatform-lib/blob/develop/rust/src/utils.rs#L13) that are was-exposable we need to use those auto-deriving macros in `utils` but this is perfectly fine for the traits. If they're only called from other rust code just the traits is totally fine and will work with the existing (de)serialization system.

rust/src/genesis/byron/config.rs
rooooooooob: Does the order matter here? Does how it's stored here impact how the hash is computed (order of some map in CBOR?) It could be not important if some order is specified in that part of code and the hashes match up, even if the order here is not what db-sync gives. I don't know enough about the genesis block to really say what should happen though.

rooooooooob (approved):
I didn't fully review the stuff that I assume was pulled from the old chain tools code (`/chain_crypto/`,`/legacy_address/` and `/genesis/*`) since it's already a part of old tooling + it's several thousand lines, but I looked at the comments you made and I went through the code that is a part of just CML (e.g. the root-level files in `rust/src` that were changed) and it looks okay. I'm assuming the brought-in code is fine since you said it works.

PR #75, opened 2022-07-01
Fix for >u64 slots in Pointer addresses

rooooooooob:
Pointer addresses with slots that don't fit in uint (u64) have been
found on testnet. To facilitate parsing these this updates Pointer to
use bigints and updates the variable nat encoding/decoding functions to
handle larger sizes.


SebastienGllmt:
Tx hash on testnet for anybody who is looking into this in the future: `95a5ef2f1121a197fc9bedb039787eb723e98825b4fd4df86c17ec0e3f4dc722`

PR #81, opened 2022-07-15
Babbage utilities

SebastienGllmt:
This PR adds in a bunch of Babbage utility functions


rust/src/serialization.rs
```diff
@@ -3698,6 +3698,63 @@ impl Deserialize for NetworkId {
     }
 }
 
+impl cbor_event::se::Serialize for Datum {
```
SebastienGllmt, line 3701: @rooooooooob I'm not sure what was up with this `TxOutputData` type. The `datum_option` type is part of the binary spec so I just replaced it with this and manually added the (de)serialization for it

PR #88, opened 2022-07-19
Simplify payment key added in component builders

SebastienGllmt:
Previously the component builders for the tx builder required knowing the vkeys ahead of time which isn't always possible. This PR sets up the framework for removing this requirement, but it's blocked at the moment because of our good friend "hacky byron support" which comes back to haunt me every single PR we make to this project.

### DIFF
--- a/rust/src/address.rs
+++ b/rust/src/address.rs
@@ -1,7 +1,7 @@
 use super::*;
 use cbor_event::{de::Deserializer, se::Serializer};
 use bech32::ToBase32;
-use crate::{ledger::common::{binary::*, value::{to_bignum, from_bignum}}, byron::{ProtocolMagic, ByronAddress}};
+use crate::{ledger::common::{value::{to_bignum, from_bignum}}, byron::{ProtocolMagic, ByronAddress}};
 
 // returns (Number represented, bytes read) if valid encoding
 // or None if decoding prematurely finished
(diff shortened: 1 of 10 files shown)


rust/src/address.rs
```diff
@@ -166,6 +167,35 @@ impl Deserialize for StakeCredential {
...
 }
 
+#[wasm_bindgen]
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd, serde::Serialize, serde::Deserialize)]
+pub struct ByronAddressAttributes(pub (crate) legacy_address::Attributes);
+
+impl ByronAddressAttributes {
+}
+
+impl cbor_event::se::Serialize for ByronAddressAttributes {
```
SebastienGllmt, line 177: The (de)serialization logic for this is all incorrect because legacy_address doesn't follow the codgen format and instead it's a frankenstein of years of piling hacks instead of properly just generating byron support

rust/src/builders/witness_builder.rs
```diff
@@ -477,6 +472,7 @@ impl TransactionWitnessSetBuilder {
                 self.add_plutus_witness_info(info);
             }
         }
+        // TODO: add in required witnesses
```
SebastienGllmt, line 475: TODO: I'll add this later, but I can't test this PR at the moment because of the Byron support issue mentioned above

PR #97, opened 2022-08-04
Handle no aux data after is_valid

SebastienGllmt:
Without this change, it seems to crash on tx `b9683e9f3dd6ff6826aa7f66614644b6c5a2d7f597039ca19fb6e2590272bc7d` on testnet with the following error

```
DeserializeError { location: Some("Transaction.auxiliary_data"), failure: CBOR(NotEnough(0, 0)) }
```

The full cbor of the transaction is added as a unit test in this PR

To be honest, I'm not sure if this is the right fix for the problem though. That being said, this piece of the parsing code is kind of hacky though and it will be replaced once we have per-era structs


rooooooooob (commented):
~I can't get that CBOR to parse anywhere (using `cbor_event` or using `cbor.me`). It reads `MAP(Len(26686))` and `ARRAY(Indefinite)` before failing.~ (edit: nevermind that's the tx hash. it parses fine now.)

I'm also not sure of which era it is allowed to encode the `null` for when there's no aux data. Even back in shelley you have:
```
transaction =
  [ transaction_body
  , transaction_witness_set
  , transaction_metadata / null
  ]
```

and now (as of Alonzo) it's
```
transaction =
  [ transaction_body
  , transaction_witness_set
  , bool
  , auxiliary_data / null
  ]
```

SebastienGllmt:
Seems this was a Pallas bug. You can find the PR that fixes this here https://github.com/txpipe/pallas/pull/159

PR #98, opened 2022-08-04
Bump wasm_bindgen + fixes for JsError

rooooooooob:
During #81 `utils.rs` was refactored into many places and some wouldn't
be using `error::JsError` which in wasm builds is a typedef for
`js_sys::JsValue` and is its own struct for rust builds (to avoid panics
in JsValue when used from non-wasm builds).

Howeve, some places couldn't see these types and would pull it in
directly from `wasm_bindgen::prelude` from the wildcare import in
`lib.rs`. This caused an issue when bumping `wasm_bindgen` as the
`from_str()` method is no longer in `js_sys::JsError` but still is in
`JsValue`. Between now and #81 this would have also led to panics if
many places when an error was returned when using CML from rust instead
of wasm.

In the future we should deprecate the direct usage of `js_sys::JsValue`
and remove our `JsError` type on rust in favor for real rust errors (not
just string wrappers).

See rustwasm/wasm-bindgen#2710

>Expands the acceptable return-position types to Result<T, E> where T: IntoWasmAbi, E: Into<JsValue>, so you can easily throw your own custom error classes imported from JS

which will greatly improve error handling from the rust side but this
should be done after the rust/wasm split.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -247,9 +247,9 @@ checksum = "1aab8fc367588b89dcee83ab0fd66b72b50b72fa1904d7095045ace2b0c81c35"
 
 [[package]]
 name = "js-sys"
-version = "0.3.51"
+version = "0.3.59"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "83bdfbace3a0e81a4253f73b49e960b053e396a11012cbd49b9b74d6a2b67062"
+checksum = "258451ab10b34f8af53416d1fdab72c22e805f0c92a1136d59470ec0b11138b2"
 dependencies = [
  "wasm-bindgen",
 ]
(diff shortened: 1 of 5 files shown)


joacohoyos:
@rooooooooob this seems to fix the memory leak issue. The only issue is with [flowgen](https://github.com/dcSpark/cardano-multiplatform-lib/blob/develop/package.json#L14). 

Removing it from the build scripts seems to make it work. If we remove that it seems to be working fine
image

rooooooooob:
Yeah I saw that flowgen thing but it seemed unrelated to any change I made so I assumed the issue existed in `develop` already although I didn't check. I just wanted to get this PR up first for testing and to see if it resolved the issue. I noticed that building also deleted the .flow types file (I didn't commit the removal since it seemed unrelated) but I am also not sure why it happened. That seems more like it'd be related to my installed `wasm-pack` version than anything here but I guess it's possible it's related to the version bump here too. I'll have to try and sort this out after I finish a different task I'm working on unless anyone else wants to look into it and needs it done in the next few days.

IIRC we might end up deprecating flow at some point anyway and just provide typescript bindings. CC: @SebastienGllmt

PR #101, opened 2022-08-08
Fix header group encoding

SebastienGllmt:
I don't know why, but these fields were skipping the array encoding and instead only serializing the content

More specifically, you can see for example in this block header cbor the operational certificate is inside an array(4), yet CML was excluding this

image

I disabled the old test since it also had the improper header encoding (presumably since the incorrect cbor was itself generated from CML). However, maybe I'm missing something

I should add that the block cbor comes from Pallas so it should be taken with a huge grain of salt, but it seems to match what I would expect from the binary spec


rooooooooob (commented):
Here's alonzo.cddl (these two fields of `header_body` stay as basic groups in other versions though, it's just `header_body` got other fields added or removed)
```
header_body =
  [ block_number     : uint
  , slot             : uint
  , prev_hash        : $hash32 / null
  , issuer_vkey      : $vkey
  , vrf_vkey         : $vrf_vkey
  , nonce_vrf        : $vrf_cert
  , leader_vrf       : $vrf_cert
  , block_body_size  : uint
  , block_body_hash  : $hash32 ; merkle triple root
  , operational_cert
  , protocol_version
  ]

operational_cert =
  ( hot_vkey        : $kes_vkey
  , sequence_number : uint
  , kes_period      : uint
  , sigma           : $signature
  )

protocol_version = (uint, uint)
```

As you can see `protocol_version` and `operational_cert` are basic/plain groups. See section 2.1 of the RFC.

The example given there are these two equivalent ways:
```
 person = {
                            age: int,
                            name: tstr,
                            employer: tstr,
                          }

                          dog = {
                            age: int,
                            name: tstr,
                            leash-length: float,
                          }

                      Figure 5: Maps with Copy/Paste

                          person = {
                            identity,
                            employer: tstr,
                          }

                          dog = {
                            identity,
                            leash-length: float,
                          }

                          identity = (
                            age: int,
                            name: tstr,
                          )

                 Figure 6: Using a Group for Factorization
```

So from what I understood when writing the codegen is you embed those fields in the type that uses them when it is a basic group.

Maybe it's a problem in Pallas of not handling basic groups?

rooooooooob:
If this is the case then there could be some other areas affected i.e.. anywhere using `pool_params` but maybe they handle those cases where it's in a group choice correctly like with `certificate` / `relay` whose variants use basic groups, as I imagine we would have noticed it if certificates were not encoded correctly. It's worth looking into though.

scarmuega:
@rooooooooob @SebastienGllmt have you checked if this is the case when parsing a raw cbor block via a direct blockfetch to the node?

rooooooooob:
@SebastienGllmt  I'm going to close this at least for now. It modifies the legacy code and while it allows parsing for babbage, it will break deserialization for all other eras. We have #246 that does this for babbage only in the current codebase.

PR #103, opened 2022-08-11
Add missing to/from JSON on Plutus types

SebastienGllmt:
Can you think a reason this was missing? Seems like just an oversight to me

### DIFF
--- a/rust/pkg/cardano_multiplatform_lib.js.flow
+++ b/rust/pkg/cardano_multiplatform_lib.js.flow
@@ -6,26 +6,6 @@
  */
 
 /**
- * @param {string} json
- * @param {number} schema
- * @returns {PlutusData}
- */
-declare export function encode_json_str_to_plutus_datum(
-  json: string,
-  schema: number
-): PlutusData;
-
-/**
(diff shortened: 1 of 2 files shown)


rooooooooob (commented):
It was either an oversight or I didn't think they were that useful at the time, but it makes sense to have them for consistency.  We still derived the traits so they can be used as part of bigger structs.

We should either remove the `to_from_json!()` macro for `ConstrPlutusData` or add it for the other plutus datum types. We have a dedicated plutus datum to/from JSON functionality based on the cardano node's format, but then we also have (inaccessible when using *directly* e.g. no `PlutusMap::from_json()) the serde format used from those derives. We needed to have the serde format as well since it's contained within other types with a to/from JSON e.g. `Transaction`.

We could try and unify the two and explicitly define the serde traits to use the cardano node's format, which probably isn't hard, but it gets a bit more complicated when it comes to the .ts definitions as I'm not sure how you'd implement `JsonSchema` in accordance manually. And you'd NEED to have that trait defined rather than being able to just handcode the .json file for the schema, since we need that trait defined in order to derive it for all containing types.

PR #105, opened 2022-08-14
Refactored metadata builder

Henry-bee:


rust/src/builders/tx_builder.rs
```diff
@@ -712,6 +717,25 @@ impl TransactionBuilder {
         Ok(())
     }
```
Henry-bee, line 719: Or should we simply remove them?

rust/src/builders/metadata_builder.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+#[wasm_bindgen]
+pub struct TxMetadataBuilder {
+    metadata: GeneralTransactionMetadata,
+}
+
+#[wasm_bindgen]
+impl TxMetadataBuilder {
+    pub fn new() -> Self {
```
SebastienGllmt, line 15: Probably it's better to have two constructors:
1. `pub fn empty()` that works like what you have now
2. `pub fn from(metadata: &GeneralTransactionMetadata)`

rust/src/builders/tx_builder.rs
SebastienGllmt: I would just remove them

rust/src/builders/metadata_builder.rs
Henry-bee: That's a great suggestion. Let me add that

PR #124, opened 2022-09-01
Sorted inputs to redeemers

MicroProofs:


rust/src/builders/tx_builder.rs
```diff
@@ -970,6 +970,14 @@ impl TransactionBuilder {
             total_collateral: self.calc_collateral_total()?,
             reference_inputs: self.reference_inputs.as_ref().map(|inputs| TransactionInputs(inputs.iter().map(|utxo| utxo.input.clone()).collect())),
         };
+
+        built.inputs.0.sort_by(|a, b| {
```
SebastienGllmt, line 974: This isn't enough because the same constraint would apply to all other tag types (mints, withdrawals, etc.)

The choice is to either sort all of these, or make that once you set the redeemers on the tx builder, you can no longer add new inputs/mints/withdrawals/certs (so basically a standalone build step)

MicroProofs: Done now withdrawals are sorted, mint was already sorted, and certs are not sorted.

rust/src/builders/tx_builder.rs
```diff
@@ -970,6 +970,26 @@ impl TransactionBuilder {
...
```
SebastienGllmt, line 974: nit: maybe add a comment saying why we sort these two fields and only these two fields

PR #125, opened 2022-09-06
now you can add a plutus input without witness if script ref present

MicroProofs:


rust/src/builders/tx_builder.rs
```diff
@@ -4508,16 +4530,18 @@ mod tests {
...
+                    &PlutusScriptWitness::from_script(
```
SebastienGllmt, line 4533: Probably it's good if we add a test for the script ref case as well

MicroProofs: sure will work on that today

SebastienGllmt (approved):

PR #127, opened 2022-09-09
feat: add reference inputs used lang to script hash calc

joacohoyos:
When using just script reference inputs no languages were passed to the calc script hash function. This caused an error in the generated safe hash not allowing the tx to be submitted.

Now, we go through both inputs and reference inputs and if there is a script in any of them we add the language

cc: @MicroProofs

### DIFF
--- a/rust/src/builders/tx_builder.rs
+++ b/rust/src/builders/tx_builder.rs
@@ -944,30 +944,40 @@ impl TransactionBuilder {
             // we make that dummy exunits set a dummy script_data_hash to ensure the tx fails if submitted to a node
             false => match has_dummy_exunit {
                 true =>  Some(ScriptDataHash::from([0u8; ScriptDataHash::BYTE_COUNT])),
-                false =>  calc_script_data_hash(
-                    &redeemers,
-                    &self.witness_builders.witness_set_builder.get_plutus_datum(),
-                    &self.config.costmdls,
-                    &Languages(
-                        self.witness_builders.witness_set_builder.scripts
-                            .values()
-                            .fold(
-                                BTreeSet::<Language>::new(),
-                                |mut langs, script| {
-                                    if let Some(lang) = &Option::<Language>::from(script) {
-                                        langs.insert(*lang);
-                                        langs
-                                    } else {
-                                        langs
-                                    }
+                false => {  
+                    let scripts = self.witness_builders.witness_set_builder.scripts.values();
+                    let mut languages = scripts.fold(BTreeSet::<Language>::new(), |mut langs, script| {
+                        if let Some(lang) = &Option::<Language>::from(script) {
+                            langs.insert(*lang);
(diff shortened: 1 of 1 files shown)


rust/src/builders/tx_builder.rs
```diff
@@ -944,28 +944,60 @@ impl TransactionBuilder {
...
+                                    |mut langs, script| {
+                                        if let Some(lang) = &Option::<Language>::from(script) {
+                                            langs.insert(*lang);
+                                            langs
+                                        } else {
+                                            langs
+                                        }
+                                    }
+                                );
+                        if let Some(reference_inputs) = self.reference_inputs.clone() {
```
SebastienGllmt, line 962: I think the indentation on everything after here is off by one

rust/src/builders/tx_builder.rs
```diff
@@ -944,28 +944,60 @@ impl TransactionBuilder {
...
+                                        if let Some(lang) = &Option::<Language>::from(&script_ref.script().0) {
+                                            langs.insert(lang.clone());
+                                             langs
+                                        } else {
+                                            langs
+                                        }
+                                    } else {
                                         langs
+                                    }
+                                }
```
SebastienGllmt, line 976: I think you can simplify the conditions like this both for ref inputs and for the inputs block that follows

```suggestion
                                    if let Some(script_ref) = &input.output.script_ref.clone() {
                                        if let Some(lang) = &Option::<Language>::from(&script_ref.script().0) {
                                            langs.insert(lang.clone());
                                        }
                                    }
                                   langs
                                }
```

rust/src/builders/tx_builder.rs
```diff
@@ -944,28 +944,60 @@ impl TransactionBuilder {
...
+                            );
+                        };
+                        self.inputs.clone().iter().fold(
+                            &mut languages,
+                            |langs, input| {
+                                if let Some(script_ref) = &input.output.script_ref.clone() {
```
SebastienGllmt, line 982: on all the places where you pattern match on an optional, you can use `as_ref()` instead of `clone()` to be slightly more efficient

rust/src/builders/tx_builder.rs
joacohoyos: @SebastienGllmt done

joacohoyos: @SebastienGllmt done

rust/src/builders/tx_builder.rs
joacohoyos: @SebastienGllmt done

rust/src/builders/tx_builder.rs
joacohoyos: @SebastienGllmt done

SebastienGllmt (approved):
LGTM

Issue #86, opened 2022-07-15
Change PlutusMap from BtreeMap to Vec


SebastienGllmt:
I noticed that Berry's fork of CML changed the PlutusMap to use a Vec instead of a BtreeMap. Not sure what the rationale for this was, but just creating an issue to track this since there may have been a good reason

SebastienGllmt:
So it sounds like it's because when constructing a transaction offchain, PlutusTx maintains insertion order whereas CML isn't because it uses a BtreeMap. Sounds like the solution may be to use a LinkedHashMap? The cddl type does use a map type though and I'm not sure what the implications of the regeneration are on this type

rooooooooob:
This was likely done due to duplicate keys. I'll close this now that we have #241

PR #130, opened 2022-09-23
Datum and redeemer insertion order

MicroProofs:
Ensure datum and redeemer order is maintained when building a transaction.

### DIFF
--- a/rust/src/builders/witness_builder.rs
+++ b/rust/src/builders/witness_builder.rs
@@ -1,4 +1,7 @@
 use std::{collections::{HashMap}, fmt::Debug};
+use itertools::Itertools;
+use linked_hash_map::LinkedHashMap;
+
 use crate::{*, ledger::common::hash::hash_plutus_data, byron::ByronAddress};
 
 use super::redeemer_builder::RedeemerWitnessKey;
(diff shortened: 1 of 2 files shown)


SebastienGllmt:
FYI this is #86 . I'm not sure about using LinkedHashMap vs a Vec since LinkedHashMap has led to some issues in the past. @rooooooooob can you sign off on this if this is okay?

rust/src/builders/witness_builder.rs
```diff
@@ -337,7 +342,9 @@ impl TransactionWitnessSetBuilder {
     }
 
     pub fn get_redeemer(&self) -> Redeemers {
-        Redeemers(self.redeemers.clone().into_values().collect())
+        Redeemers(self.redeemers.clone().iter().map(|i|{
```
rooooooooob, line 345: I don't think this clone is necessary anymore since `iter()` should give you refs to iterate over which you are cloning in that map anyway. Same with the one above in `get_plutus_datum`.

rooooooooob (commented):
@SebastienGllmt this won't address #86 fully though as `PlutusMap` is still a `BTreeMap`. Although I guess that is more for deserialization and such. This wouldn't cause an issue with the order conflicting in the built txs right? Since those are used as lists? Does this address the reason why in #86 they had changed it? Was it for building txs?

Could you refresh my mind on what issues `LinkedHashMap` caused? (bug-wise). I can only remember it causing issues due to `BTreeMap` automatically being canonical form so if we needed to serialize canonically (e.g. tx hashes) switching this out would break it. Or maybe traits not being implemented for it e.g. the JSON schema ones or CBOR ones which is why we have a wrapper around it in the latest version of codegen to implement those. Neither of those should be relevant here so I think it's okay, assuming this addresses the root reason why in the fork they changed it.

SebastienGllmt:
Yeah, I think the issues with LinkedHashMap were:
- issues preserving encodings
- requires manual conversion to canonical form
- requires manual to/from json implementations

rooooooooob:
It should *fix* the preserving encodings though, not cause it. That's why we've been using it in places. `BTreeMap` didn't since it just orders lexicographically. (which may or may not be canonical order everywhere, since canonical order is canonical order of the serialized bytes but should be fine for most simple keys).

If this covers the root reason why the fork changed it I'm fine with this change, although the minor nit would be those clones are unnecessary afaik. A full replacement to `LinkedHashMap` to maintain order in `PlutusMap` should happen anyway after the regeneration. I thought we had already changed it, but it looks like I just stored the original bytes of the entire datum instead to cover all encoding issues in Emurgo/cardano-serialization-lib#317

MicroProofs:
Cool I'll change the clones

rooooooooob (approved):
LGTM

PR #137, opened 2022-10-11
Rust/WASM initial regeneration

rooooooooob:
This contains the direct code after generating off of the binary spec in `specs/babbage/`.

We will now be working out of these two crates in the `develop` branch, but they will just be sitting there unused until we do the final switch over where we delete the root `rust` crate and have everything moved over to `rust/core` and `rust/wasm`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,9 @@ rust/target
 .idea
 rust/.idea
 binaryen/
+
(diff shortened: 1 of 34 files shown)


rooooooooob:
This finishes #141

gostkin (approved):

PR #143, opened 2022-10-21
Crypto / address refactoring

rooooooooob:
Moving over existing hand-written (not code-generated) code from the original crate over to the new core crate. Includes refactoring that functionality to match the new format for the regeneration (native rust API with separate WASM bindigns) + preserving CBOR formats + actual rust errors.


rust/core/Cargo.toml
```diff
@@ -13,3 +13,46 @@ derivative = "2.2.0"
...
+# These can be removed if we make wasm bindings for ALL functionality here.
+# This was not done right now as there is a lot of existing legacy code e.g.
+# for Byron that might need to be used from WASM and might not.
+# We can remove this dependency when that is decided.
+#
+# The other use-case here is enums. Without this two enums would need to be defined
+# despite wasm_bindgen supporting C-style enums (with non-negative values) 100%
+# This could possibly be resolved with macros but maybe not.
+
+# non-wasm
```
rooooooooob, line 43: It seems we will have to use the full wasm_bindgen not the noop_proc_macro here as in WASM builds the `core` crate will get compiled using `noop_proc_macro` even when it's only used as a dependency by the `wasm` crate in a WASM build.

rust/core/src/crypto.rs
```diff
@@ -517,43 +437,38 @@ impl From<key::EitherEd25519SecretKey> for PrivateKey {
     }
 }
 
-#[wasm_bindgen]
 impl PrivateKey {
     pub fn to_public(&self) -> PublicKey {
         self.0.to_public().into()
     }
 
-    pub fn generate_ed25519() -> Result<PrivateKey, JsError> {
```
rooooooooob: I have no idea why this was a `Result`. It all chains off an initial `Ok()`

rust/core/src/lib.rs
```diff
@@ -64,32 +76,29 @@ pub type Mint = OrderedHashMap<PolicyId, OrderedHashMap<AssetName, Int64>>;
...
+pub type ProposedProtocolParameterUpdates = OrderedHashMap<GenesisHash, ProtocolParamUpdate>;
 
 pub type RewardAccounts = Vec<RewardAccount>;
 
-pub type ScriptDataHash = Hash32;
-
+// TODO: this should NOT be generating an alias! (it's both tagged and a .cbor script)
+// we should investigate cddl-codegen and open an issue there if it's still happening
+// (top-level .cbor was recently being worked on and this is tagged but even then it should NOT be bytes here)
 pub type ScriptRef = Vec<u8>;
```
rooooooooob, line 100: Issue: [cddl-codegen#117](https://github.com/dcSpark/cddl-codegen/issues/117)

rust/core/src/serialization.rs
```diff
@@ -187,57 +203,69 @@ pub trait SerializeEmbeddedGroup {
     ) -> cbor_event::Result<&'a mut Serializer<W>>;
 }
 
+pub trait Deserialize {
+    fn deserialize<R: BufRead + Seek>(
```
rooooooooob, line 207: These changes can be backported to cddl-codegen so that we can generate the appropriate code for the WASM wrappers since traits don't work in WASM.

Do we want `to_bytes()` and `to_cbor_bytes()` for all WASM wrappers? We only had `to_bytes()` before but I felt it was potentially confusing as in the old code this was equivalent to the CBOR binary spec *most* of the time, but not always. This way it should be more obvious.

rust/core/src/crypto.rs
```diff
@@ -937,6 +786,10 @@ macro_rules! impl_hash_type {
             fn serialize<'se, W: std::io::Write>(&self, serializer: &'se mut Serializer<W>, force_canonical: bool) -> cbor_event::Result<&'se mut Serializer<W>> {
                 serializer.write_bytes_sz(&self.data, self.encoding.to_str_len_sz(self.data.len() as u64, force_canonical))
             }
+
+            fn to_bytes(&self) -> Vec<u8> {
```
rooooooooob, line 790: I didn't put this in every place that has a to/from raw bytes yet. There's a couple places that don't have it but they'll need to be added. Question: Should this **replace** to_raw_bytes()? i.e. to_bytes() will be raw bytes when possible (hashes, addresses, etc) and CBOR bytes if not, or should there be all 3 in cases where there is a raw bytes representation? e.g. `to_cbor_bytes()`, `to_bytes()`, `to_raw_bytes()` to be maximally explicit. Should all 3 be exposed to WASM?

rooooooooob:
We'll still need to split up the crypto (and maybe address?) into separate crates later. I'm imagining that the chain (or babbage or however we organize it) crate will just pull those in as dependencies. However, in some cases these are also used on-chain in some cases e.g. `Vkey`, `RewardAddress`, etc. Would we want to expose the on-chain representation details (how it's encoded, serialization traits, etc) to the general `crypto` crate? Or have the crypto be 100% on its own with the crypto/address related things that have on-chain representations be wrappers around it e.g. `Ed25519KeyHash { keyhash: crypto::Ed25519KeyHash, encoding: ...}` ?

rust/core/src/crypto.rs
gostkin: @SebastienGllmt what do you think?

rust/core/Cargo.toml
```diff
@@ -13,3 +13,46 @@ derivative = "2.2.0"
...
+bech32 = "0.7.2"
+hex = "0.4.0"
+itertools = "0.10.1"
+getrandom = { version = "0.2.3", features = ["js"] }
+rand = "0.8.5"
+fraction = "0.10.0"
+base64 = "0.13"
+num-bigint = "0.4.0"
+num-integer = "0.1.45"
+#rand_os = "0.1"
```
gostkin, line 30: should we remove that?

rust/core/src/address.rs
```diff
@@ -1,31 +1,1041 @@
-#[derive(Clone, Debug, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
-pub struct Address {
-    #[serde(skip)]
-    pub encodings: Option<AddressEncoding>,
+use super::*;
+use schemars::JsonSchema;
+use cbor_event::{de::Deserializer, se::Serializer};
+use bech32::ToBase32;
+//use crate::byron::{ProtocolMagic, ByronAddress, ByronAddressError};
```
gostkin, line 5: do we need that here?

rust/core/src/address.rs
```diff
@@ -1,31 +1,1041 @@
...
+    CBOR(#[from] DeserializeError),
+}
+
+#[derive(Debug, Clone, Eq, Ord, PartialEq, PartialOrd)]
+pub enum Address {
+    Base(BaseAddress),
+    Ptr(PointerAddress),
+    Enterprise(EnterpriseAddress),
+    Reward(RewardAddress),
+    //Byron(ByronAddress),
```
gostkin, line 96: will we uncomment when byron is regenerated?

gostkin:
why we need to regenerate the wasm code again? what's changed?

rust/core/src/crypto.rs
SebastienGllmt: If we have 3 different functions, I would avoiding naming one of them a generic `to_bytes` because otherwise people will use it without understanding the difference between the 3

rooooooooob:
We need to regenerate it due to dcSpark/cddl-codegen#114

rust/core/src/address.rs
rooooooooob: We will when we bring over the byron module.

rust/core/src/crypto.rs
rooooooooob: Do we even want 3 in that case? Or just `to_cbor_bytes()` for everything and then additionally `to_raw_bytes()` for the types that have a raw bytes representation? (crypto stuff, addresses)

rust/core/src/lib.rs
gostkin: should be resolved already

SebastienGllmt (approved):

PR #150, opened 2022-11-02
Change PR checks trigger strategy

gostkin:
It makes sense to run CI when PR is open. Otherwise we can't verify it works and compiles unless we submit review / run locally. Before reviewing it makes sense quite often to see that checks have passed.

This relates to dependabot updates as well: good to see whether CI passes before reviewing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ rust/core/Cargo.lock
 rust/wasm/target/**
 rust/wasm/Cargo.lock
 rust/json-gen-split/target/**
+tools/metadata-cddl-checker/Cargo.lock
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -11,3 +11,4 @@ rust/core/Cargo.lock
 rust/wasm/target/**
 rust/wasm/Cargo.lock
 rust/json-gen-split/target/**
+tools/metadata-cddl-checker/Cargo.lock
```
rooooooooob, line 14: Shouldn't this not be here since this a tool people use as an application not as a library used by other rust programs?

gostkin: currently there's no .lock in that folder, we either should ignore the file in git or generate and add

PR #152, opened 2022-11-11
Fix Script serialization

SebastienGllmt:
This fixes an issue in `Script` serialization where we were accidentally using tags instead of unsigned ints.

This was causing errors in round-trip serialization of the `Script` type


rooooooooob (approved):
LGTM

PR #153, opened 2022-11-30
Workspaces split

rooooooooob:
Splits the existing `rust/core` and `rust/wasm` into multiple crates at a root workspace. This also adds in the `cip25-lib` repo.


Cargo.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
+
+# this is for the new crate structure. The legacy code (current CML) still resides in the `rust` directory.
+members = [
+    "core",
```
rooooooooob, line 5: We'll still need to figure out what to do with the JSON schema generating stuff as we currently have multiple in the project now with CIP25.

chain/Cargo.toml
```diff
@@ -0,0 +1,54 @@
...
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+cardano-multiplatform-lib-core = { "path" = "../core" }
+cardano-multiplatform-lib-crypto = { "path" = "../crypto" }
+cbor_event = "2.2.0"
+linked-hash-map = "0.5.3"
```
rooooooooob, line 13: Some of the dependencies will need to be cleared up and such but we can do that after we finalize the workspace structure.

core/src/serialization.rs
```diff
@@ -0,0 +1,259 @@
...
+pub trait Serialize {
+    fn serialize<'a, W: Write + Sized>(
+        &self,
+        serializer: &'a mut Serializer<W>,
+        force_canonical: bool,
+    ) -> cbor_event::Result<&'a mut Serializer<W>>;
+
+    /// Bytes of a structure using the CBOR bytes as per the CDDL spec
+    /// which for foo = bytes will include the CBOR bytes type/len, etc.
+    fn to_cbor_bytes(&self, force_canonical: bool) -> Vec<u8> {
```
rooooooooob, line 179: One thing we have different is that CIP25 was generated without encoding preservation so there's this `force_canonical` concept on serialization. It's ignored by the auto-implementation of `Serialize` if `cbor_event::Serialize` is implemented, which is the trait that CIP25 uses. Is that the behavior we wanted? CIP25 was generated without the encoding preservation since it didn't really need it, from what I thought you wouldn't be storing the CBOR here and you'd be using it more for parsing existing metadata. The `force_canonical` is there since it's used for hardware wallets for encoding tx bodies but might not be too useful in other places.

Cargo.toml
gostkin: could you share a little more context please? what's the problem? seems like I don't have full picture

cip25-wasm/src/lib.rs
```diff
@@ -0,0 +1,833 @@
+use wasm_bindgen::prelude::*;
+
+// TODO: remove after regen
```
gostkin, line 3: should we regenerate?

Cargo.toml
rooooooooob: For the WASM build in legacy CML we have as part of a build script a rust program in `rust/json-gen` that outputs JSON schemas that define the JSON format of on-chain data. We have a similar script in `json-gen-split`. We also have that with CIP-25 as well since we used the JSON functionality of cddl-codegen too. The program outputs those schemas then the rest of the build scripts piece those together (see the `scripts/` folder + `package.json` for how the scripts/programs are composed) into the final output `.ts` file output by wasm-bindgen / wasm-pack to represent the types from JS/typescript. We're going to have to update the build scripts and such now that we'll have multiple wasm builds as well.

cip25-wasm/src/lib.rs
rooooooooob: We will regenerate very soon. After we finalize this crate structure + maybe add in a quick feature to cddl-codegen to mark a type as existing without having to define it to help with generating code that is defined elsewhere (e.g. address, crypto, stuff in other crates) + maybe a parameter/config to set crate names now that we have multiple crates and the `core` crate name in cddl-codegen from before can be confusing now that it'd actually be in `chain` and `cip25`  etc whereas `core` is a different non-generated crate.

core/src/serialization.rs
gostkin: if it was generated with the encoding preservation, would this return just the original rep? canonical means https://www.rfc-editor.org/rfc/rfc7049#section-3.9 canonical cbor from the standard, right?

gostkin:
cargo test fails if we run it from workspace folder

```
cannot find attribute `quickcheck` in this scope
   --> crypto/src/chain_crypto/algorithms/ed25519_derive.rs:116:7
    |
116 |     #[quickcheck]
```

gostkin:
let's merge #150 , so we can see how the things are with checks before review, @rooooooooob ?

core/src/serialization.rs
rooooooooob: So the stuff in `chain` right now if you pass in `false` will return whatever format was used when it was deserialized, or canonical if there was no format (e.g. was created from code right now not deserialized). If you pass in `true` it will turn it into canonical CBOR regardless of how it was created.

In CIP25 whatever you pass in it'll end up being canonical (afaik). So if you deserialize something it won't preserve the format, which was fine before when it was on its own since you're not going to be storing anything here, as this is more just for parsing without reencoding ever.

We could regenerate it with preserve-encdoings=true but that won't work with how it's used in Cardano as the CDDL file for the spec is actually a subset of the input bytes since it only parses the subset of a given tx metadata that conforms to the spec. But to make round-tripping consistent you'd need to figure out a way to remember the encodings of all other parts of the metadata NOT involved in the cip25 spec. We could potentially look at modifying cddl-codegen to figure out a way to do that but I'm not sure it's worth the time or complexity it adds to the crate. It would be nice if it was preserve-encodings for consistency now that we have a mono-repo and it's a part of CML, but on the other hand we could just document that it's not for round-tripping unless we think this is something important. @SebastienGllmt

rooooooooob:
I pushed another commit that moves the on-chain crypto primitives into the `chain` crate in a `crypto` module. I kept base (not caring about encodings) versions of these in the `crypto` crate with `ChainCrypto<T>` wrappers around these in the `chain/crypto` module that take care of the encoding information relevant to their on-chain-ness.

chain/src/crypto.rs
```diff
@@ -0,0 +1,372 @@
...
+    }
+}
+
+impl<'de, T: cml_crypto::RawBytesEncoding> serde::de::Deserialize<'de> for ChainCrypto<T> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where
+    D: serde::de::Deserializer<'de> {
+        let s = <String as serde::de::Deserialize>::deserialize(deserializer)?;
+        T::from_raw_hex(&s)
+            .map(Into::into)
+            .map_err(|_e| serde::de::Error::invalid_value(serde::de::Unexpected::Str(&s), &"hex bytes for signature"))
```
rooooooooob, line 270: Oops copy-paste error. I'll change this message so it doesn't say signature (since this could be a hash too).

chain/src/lib.rs
```diff
@@ -200,9 +188,9 @@ pub mod certs;
 pub use certs::*;
 
 
-pub mod crypto;
+//pub mod crypto;
```
gostkin, line 191: can be removed , we have public exposure above

gostkin:
looks good

core/src/serialization.rs
SebastienGllmt: Yeah, handling this kind of subset operation is a ticket we have no plans to address at the moment

rooooooooob: So I was thinking about two options then:
a) Leave it how it is and document that the cip-25 crate doesn't preserve encodings and should not be used for round-tripping.
b) get rid of `Serialize::to_cbor_bytes(force_canonical: bool)` and instead make 2 other traits `SerializeOriginal::to_original_cbor_cbytes()` and `SerializeCanonical::to_canonical_cbor_bytes()` but it's feeling awfully pedantic now. In this case `cip-25` would only implement the canonical one and `chain` would implement both. Although this doesn't explicitly cover the fact that `cip-25` ignores fields not part of the spec either.

or something along those lines

Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+    "crypto",
+    "cip25",
+    "cip25-wasm",
+    "wasm"
+]
+
+# exclude old crate structure to avoid error in it
+exclude = [
```
gostkin, line 13: how the pipeline will be working? will we have one for old and one for new structure? we need to verify both old and new crates are compiling and working correctly

gostkin (approved):

PR #156, opened 2022-12-05
Use custom format to serialize the json

gostkin:
- [x] Impl serialize
- [x] Impl deserialize

### DIFF
--- a/rust/src/json_serialize.rs
+++ b/rust/src/json_serialize.rs
@@ -1,10 +1,16 @@
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, VecDeque};
+use std::fmt::{Display, Formatter};
+use std::iter::FromIterator;
+
 use itertools::Itertools;
-use serde::{Deserialize, Deserializer};
-use serde_json::{Map, Number};
-use crate::{DeserializeError, JsError};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::ser::Error;
+use unicode_segmentation::UnicodeSegmentation;
+
+use crate::JsError;
 use crate::ledger::common::value::BigInt;
 
+#[derive(Debug, Clone, Eq, PartialEq)]
 pub enum Value {
     Null,
     Bool(bool),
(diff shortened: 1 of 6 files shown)


rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+                (true, Some(JsonToken::ArrayEnd))
+            }
+            ":" => {
+                (true, Some(JsonToken::Colon))
+            }
+            "," => {
+                (true, Some(JsonToken::Comma))
+            }
+            _ => {
+                let splitted: Vec<&str> = char.split_whitespace().into_iter().collect();
```
gostkin, line 211: that part of the code is not beautiful, sorry about that. that was the easiest way to get rid of whitespaces

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+    Bool(bool),
+    Number(BigInt),
+    String(String),
+    Array(Vec<Value>),
+    Object(BTreeMap<String, Value>),
+}
+
+impl Display for Value {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        f.serialize_str(self.clone().to_string().map_err(|err| std::fmt::Error::custom(format!("Can't serialize {:?}", err)))?.as_str())
```
rooooooooob, line 33: Maybe this is a dumb question, but how does this work? We've got `use std::fmt::{Display, Formatter};` and `Display` uses that one, so where do we get `serialize_str` from? I see it's in `serde::fmt::Formatter`.

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+    }
+}
+
+fn tokenize_string(string: String) -> Vec<JsonToken> {
+    let mut tokens = Vec::<JsonToken>::new();
+    let mut current_string: String = String::new();
+    for char in string.graphemes(true) {
+        let (reset_string, token) = match char {
+            "\"" => {
+                if !current_string.is_empty() && current_string.clone().graphemes(true).last().clone().unwrap() == "\\" {
```
rooooooooob: Are the clones necessary here to call `graphemes()`? The crate seems to take a reference although it's implemented on `str` not `String`.

rust/src/json_serialize.rs
rooooooooob: Can we strip the whitespace in a single pass beforehand instead to simplify things?

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+    },
+
+    LeftQuotedString {
+        raw: String,
+    },
+
+    ParsedValue {
+        value: Value,
+    },
+    ParsedKey {
```
rooooooooob, line 61: I would add a comment that this contains the `:` token inside it I guess just to make it completely clear.

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+            serde_json::Value::Array(serde_arr).to_string()
+        )
+    }
+
+    fn generate_array_of_primitives() -> Vec<Value> {
+        let mut cml_arr = vec![Value::Null];
+        cml_arr.extend(vec![true, false].into_iter().map(|b| Value::Bool(b)));
+        cml_arr.extend(vec![0, u64::MAX].into_iter().map(|integer| Value::Number(BigInt::from(integer))));
+        cml_arr.extend(vec![0, i64::MAX, i64::MIN].into_iter().map(|integer| Value::Number(BigInt::from(integer))));
+        cml_arr.extend(vec!["supported_string", ""].into_iter().map(|str| Value::String(String::from(str))));
```
rooooooooob, line 637: Should we have a unicode test?

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+    fn generate_array_of_primitives_serde_json() -> Vec<serde_json::Value> {
+        let mut serde_arr = vec![serde_json::Value::Null];
+        serde_arr.extend(vec![true, false].into_iter().map(|b| serde_json::Value::Bool(b)));
+        serde_arr.extend(vec![0, u64::MAX].into_iter().map(|integer| serde_json::Value::from(integer)));
+        serde_arr.extend(vec![0, i64::MAX, i64::MIN].into_iter().map(|integer| serde_json::Value::from(integer)));
+        serde_arr.extend(vec!["supported_string", ""].into_iter().map(|str| serde_json::Value::from(String::from(str))));
+        serde_arr
+    }
+
+    fn generate_array_of_primitives_with_unsupported() -> Vec<Value> {
```
rooooooooob, line 651: A comment might help to someone reading this later without knowing why we did this as to why it's called unsupported.

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
...
+        run_cases(cases);
+    }
+
+    #[test]
+    fn mix() {
+        let cases = vec![
+            generate_array(generate_objects()),
+            generate_array(generate_arrays()),
+            generate_object(generate_arrays()),
+            generate_object(generate_arrays()),
```
rooooooooob: Was this meant to be `generate_object(generate_objects()),` to cover all 4 mixes?

rust/src/json_serialize.rs
```diff
@@ -0,0 +1,996 @@
+use std::collections::{BTreeMap, VecDeque};
```
rooooooooob, line 1: We might want to document why we have this instead of using the serde dependency that we use in other parts of the codebase.

rooooooooob (commented):
Looks good in general just some small comments.

rust/src/json_serialize.rs
gostkin: unfortunately no, technically whitespace is a valid symbol in strings

rust/src/json_serialize.rs
gostkin: basically if you can provide a string it will be interpreted / serialized as a string. so I use my `to_string()` to convert to string and then use the standard serialize_str. Not sure if it's the best approach, but it should work at least

rust/src/json_serialize.rs
gostkin: yes, there're some https://github.com/dcSpark/cardano-multiplatform-lib/blob/53169bc7e787099a694dbdc72eaa4d94dd7a958e/rust/src/json_serialize.rs#L753

gostkin: if you have an idea for better test - pls tell

rust/src/json_serialize.rs
rooooooooob: Oh yeah of course that makes sense.

rooooooooob (approved):
LGTM

PR #159, opened 2022-12-22
Cip36 lib

rooooooooob:
Main CIP36 work on the rust side.  This includes:
* Initial generation from `specs/cip36.cddl`
* clean-up of code (rust+WASM)
* convenience functions for converting to/from Metadata. Metadata was refactored into `core` to help with this (more so with CIP-25 as we need `chain` here anyway). (Rust only)
* crypto integration + creation of hash to sign under CIP format
* test against cip36 test vectors (weighted test doesn't pass but this seems to be due to inconsistencies with the test vectors or the CDDL spec)


chain/Cargo.toml
```diff
@@ -43,7 +43,7 @@ thiserror = "1.0.37"
 
 # wasm
 #[target.'cfg(all(target_arch = "wasm32", not(target_os = "emscripten")))'.dependencies]
-wasm-bindgen = { version = "=0.2.82", features = ["serde-serialize"] }
+wasm-bindgen = { version = "=0.2.83", features = ["serde-serialize"] }
```
rooooooooob, line 46: This version is required for using serde wasm bindgen dependency which we added a bit ago to avoid potential cyclical dependency issues.

crypto-wasm/src/lib.rs
```diff
@@ -166,6 +174,12 @@ impl From<Bip32PublicKey> for core_crypto::Bip32PublicKey {
     }
 }
 
+impl AsRef<core_crypto::Bip32PublicKey> for Bip32PublicKey {
+    fn as_ref(&self) -> &core_crypto::Bip32PublicKey {
```
rooooooooob, line 178: I'll make `cddl-codegen` spit these out for all types since they're a lot more needed now for us. Before we could do `foo.0` to avoid cloning just to read a wasm parameter's inner rust value, but now they can be in other modules and we can't `pub` it as that messes wasm_bindgen as it wants to expose it to wasm too. This should be cleaner than doing `.0` too.

gostkin (commented):
cip36 wasm doesn't compile on my side

gostkin:
``` 
= note: rust-lld: error: duplicate symbol: ed25519signature_from_raw_bytes
          >>> defined in /Users/gostkin/CLionProjects/cardano-multiplatform-lib/target/wasm32-unknown-unknown/release/deps/cardano_multiplatform_lib_cip36_wasm.cardano_multiplatform_lib_cip36_wasm.62a3b95f-cgu.1.rcgu.o
          >>> defined in /Users/gostkin/CLionProjects/cardano-multiplatform-lib/target/wasm32-unknown-unknown/release/deps/libcardano_multiplatform_lib_crypto_wasm.rlib(cardano_multiplatform_lib_crypto_wasm.cardano_multiplatform_lib_crypto_wasm.9a5ad4cd-cgu.0.rcgu.o)
          
          rust-lld: error: duplicate symbol: __wbindgen_describe_ed25519signature_from_raw_bytes
          >>> defined in /Users/gostkin/CLionProjects/cardano-multiplatform-lib/target/wasm32-unknown-unknown/release/deps/cardano_multiplatform_lib_cip36_wasm.cardano_multiplatform_lib_cip36_wasm.62a3b95f-cgu.1.rcgu.o
          >>> defined in /Users/gostkin/CLionProjects/cardano-multiplatform-lib/target/wasm32-unknown-unknown/release/deps/libcardano_multiplatform_lib_crypto_wasm.rlib(cardano_multiplatform_lib_crypto_wasm.cardano_multiplatform_lib_crypto_wasm.9a5ad4cd-cgu.0.rcgu.o)
          
          rust-lld: error: duplicate symbol: __wbg_ed25519signature_free
          >>> defined in /Users/gostkin/CLionProjects/cardano-multiplatform-lib/target/wasm32-unknown-unknown/release/deps/cardano_multiplatform_lib_cip36_wasm.cardano_multiplatform_lib_cip36_wasm.62a3b95f-cgu.1.rcgu.o
          >>> defined in /Users/gostkin/CLionProjects/cardano-multiplatform-lib/target/wasm32-unknown-unknown/release/deps/libcardano_multiplatform_lib_crypto_wasm.rlib(cardano_multiplatform_lib_crypto_wasm.cardano_multiplatform_lib_crypto_wasm.9a5ad4cd-cgu.0.rcgu.o)
          

warning: `cardano-multiplatform-lib-cip36-wasm` (lib) generated 4 warnings
error: could not compile `cardano-multiplatform-lib-cip36-wasm` due to previous error; 4 warnings emitted
Error: Compiling your crate to WebAssembly failed
Caused by: failed to execute `cargo build`: exited with exit status: 101
  full command: "cargo" "build" "--lib" "--release" "--target" "wasm32-unknown-unknown"
```

chain/src/address.rs
```diff
@@ -1,14 +1,40 @@
-use super::*;
 use schemars::JsonSchema;
```
rooooooooob, line 1: These changes were mostly due to trying to move address to its own module to avoid pulling in `chain` but that isn't as easy to do as it relies on multiple things in `chain`. I kept these small changes as it will make it easier to try and move it out if we ever need to do that again to optimize space or for whatever other reason.

chain/src/auxdata.rs
```diff
@@ -0,0 +1,77 @@
+/// TX auxiliary data
+use super::*;
```
rooooooooob, line 2: This is the old `chain/src/metadata.rs` but with the metadata stuff removed and moved to `core/src/metadata.rs`

cip36/src/lib.rs
```diff
@@ -0,0 +1,602 @@
...
+    /// Add to an existing metadata (could be empty) the full CIP36 deregistration metadata
+    pub fn add_to_metadata(&self, metadata: &mut Metadata) -> Result<(), DeserializeError> {
+        let dereg_metadatum = TransactionMetadatum::from_cbor_bytes(&self.key_deregistration.to_original_cbor_bytes())?;
+        metadata.insert(KEY_DEREGISTRATION_LABEL, dereg_metadatum);
+        let witness_metadatum = TransactionMetadatum::from_cbor_bytes(&self.deregistration_witness.to_original_cbor_bytes())?;
+        metadata.insert(DEREGISTRATION_WITNESS_LABEL, witness_metadatum);
+        Ok(())
+    }
+
+    // these are not implementing Serialize/Deserialize as we do not keep track of the rest of the encoding metadata
```
rooooooooob, line 155: I didn't like them implementing the same `Serialize` and having it provide `to_original_cbor_bytes()` when that isn't quite true as it might be a subset. I don't think this solution is particularly pretty but it should lead to less possible confusion. You really shouldn't be round-tripping this anyway, but instead using it for either:
1) creating the metadata in the first place
2) parsing on-chain stuff for read-only purposes that came from `Metadata` at some point (use the `Metadata` object for the round-tripping if you care about encodings)

cip36/src/lib.rs
```diff
@@ -0,0 +1,602 @@
...
+            encodings: None,
+        }
+    }
+
+    /// Create bytes to sign to make a `DeregistrationWitness` from.
+    ///
+    /// # Arguments
+    ///
+    /// * `force_canonical` - Whether to encode the inner registration canonically. Should be true for hardware wallets and false otherwise.
+    pub fn hash_to_sign(&self, force_canonical: bool) -> cbor_event::Result<Vec<u8>> {
```
rooooooooob, line 321: Do we want to have a function that combines this all in one step with a key from `crypto` to directly produce the witness? I left this as-is in case people are using it as part of other libs or need to pass it elsewhere (is this ever needed? e.g. dApp connector or whatever or would this only ever be done in one step all in the wallet?)

core/src/lib.rs
```diff
@@ -3,5 +3,120 @@ pub mod error;
...
+pub type Epoch = u64;
+
+pub type Slot = u64;
+
+pub type TransactionIndex = u16;
+
+pub type CertificateIndex = u64;
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize, schemars::JsonSchema, Derivative)]
+#[derivative(Eq, PartialEq, Ord="feature_allow_slow_enum", PartialOrd="feature_allow_slow_enum", Hash)]
```
rooooooooob, line 32: pulled from `chain/src/lib.rs` since it's used in other places like metadata and seems general enough (supporting CDDL's `int` type)

core/src/serialization.rs
```diff
@@ -176,18 +180,21 @@ pub trait Serialize {
 
     /// Bytes of a structure using the CBOR bytes as per the CDDL spec
     /// which for foo = bytes will include the CBOR bytes type/len, etc.
-    fn to_cbor_bytes(&self, force_canonical: bool) -> Vec<u8> {
+    /// This gives the original bytes in the case where this was created
+    /// from bytes originally
+    fn to_original_cbor_bytes(&self) -> Vec<u8> {
         let mut buf = Serializer::new_vec();
```
rooooooooob, line 186: having two functions seemed to be the consensus from the meetings. is this what we want to stay by? it's the most descriptive I think. I'm not sure how much of a deal it'd be space-wise for the JS lib if every single object has an extra function though.

rooooooooob:
I'm marking this ready to review without the WASM bindings fully done as that makes more sense to finalize those after feedback on the rust code first. We can also do that as a later PR as we'll want to get this merged sooner rather than later as some of the refactoring done here will be needed to finish up CIP25.

cip36/src/serialization.rs
```diff
@@ -0,0 +1,550 @@
...
+                }),
+            })
+        })().map_err(|e| e.annotate("KeyDeregistration"))
+    }
+}
+
+impl Serialize for KeyRegistration {
+    fn serialize<'se, W: Write>(&self, serializer: &'se mut Serializer<W>, force_canonical: bool) -> cbor_event::Result<&'se mut Serializer<W>> {
+        serializer.write_map_sz(self.encodings.as_ref().map(|encs| encs.len_encoding).unwrap_or_default().to_len_sz(4 + if self.voting_purpose != 0 || self.encodings.as_ref().map(|encs| encs.voting_purpose_default_present).unwrap_or(false) { 1 } else { 0 }, force_canonical))?;
+        let (legacy_format, should_include_voting_purpose) = match self.delegation {
```
rooooooooob, line 309: this bit was manually modified. I'm not sure what the desired behavior should even be. Under CDDL `.default` allows you to omit it if it's the default value. The test vectors have it present even as 0, but only in the weighted example. It shouldn't be there if it's canonical CBOR (e.g. for HW wallets I'm guessing?) so we might want an easier way (besides modifying the encodings struct manually). Should it ALWAYS be omitted in legacy examples? The CDDL spec on its own lets you do whatever, but I got the feeling that the legacy version was only for backwards compatability with CIP-15 i.e. it shouldn't be there unless you're using weighted, and even then it might be optional?

core/src/serialization.rs
gostkin: should be fine imo

chain/src/address.rs
gostkin: will the address be a separate crate in the end?

rooooooooob: maybe if `chain` becomes more bloated, which could happen once we add in the other eras. The biggest contribution to both is the `crpyto` crate so the extra bit added by `chain` didn't seem significant for crates like `cip36` that needed both addresses + crypto, and addresses need hashes (in `crypto`) anyway. For flint the size differences we're talking about aren't really significant (and we'll need all crates anyway), so it might matter more for dapps that just need 1 part, but for now I don't think we need to worry about it.

cip36/src/serialization.rs
gostkin: logically it looks like it's made for backwards compatibility, but it's cardano, if it turns out one day this thing can be used with legacy addresses, it can bring some trouble. @SebastienGllmt pls tell what you think when you have time

gostkin (approved):
overall lgtm, let's unblock further cip25 stuff

PR #164, opened 2023-01-13
CIP25 updates + add core/wasm crate + regen core/metadata

rooooooooob:
More work on CIP25 to add tests/improve API

Add a core-wasm crate for bindings over core. This is meant to be consumed from our other WASM crates i.e. cip25-wasm/cip36-wasm

Regenerate core/metadata + core-wasm/metadata

Add some documentation to CIP25

### DIFF
--- a/cip25-wasm/README.md
+++ b/cip25-wasm/README.md
@@ -1,3 +1,47 @@
 # cardano-multiplatform-lib-cip25
 
 Multiplatform SDK for CIP25 (Cardano NFT Metadata)
+
+This can be used for both parsing and creating CIP25-compatible metadata.
+
+Parsing can either be done directly from metadata bytes:
+
+```javascript
+let schema = wasm.CIP25.CIP25Metadata.from_bytes(Buffer.from("a11902d1a26464617461a144baadf00da144cafed00da5646e616d656d4d65746164617461204e616d656566696c657382a3637372636473726331646e616d656966696c656e616d6531696d65646961547970656966696c657479706531a3637372636473726332646e616d656966696c656e616d6532696d65646961547970656966696c65747970653265696d6167657821687474733a2f2f736f6d652e776562736974652e636f6d2f696d6167652e706e67696d656469615479706567696d6167652f2a6b6465736372697074696f6e776465736372697074696f6e206f662074686973204e46546776657273696f6e02", "hex"));
+
+// the above CBOR hex is for a V2 CIP25 schema. You should check which type it is first
+// as as_label_metadata_v2() will return None (undefined) if it's a v1 schema
+let v2Schema = schema.key_721().as_label_metadata_v2().data();
+let policies = v2Schema.keys();
+for (var i = 0; i < policies.len(); ++i) {
+  let policy = policies.get(i);
+  let assets = v2Schema.get(policy);
+  let assetNames = assets.keys();
+  for (var j = 0; j < assetNames.len(); ++j) {
+    let assetName = assetNames.get(j);
+    let details = assets.get(assetName);
+    console.log(`CIP25 NFT ${policy.toString("hex")}.${asset_name.toString("hex")} found:`);
+    console.log(`  name: ${details.name().to_str()}`);
+    console.log(`  image: ${details.image().to_string()}`);
+    let mediaType = details.media_type();
+    if (mediaType != null) {
+      console.log(`  media_type: ${mediaType.to_str()}`);
+    }
+    let description = details.media_type();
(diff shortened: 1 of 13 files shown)


cip25-wasm/README.md
```diff
@@ -1,3 +1,47 @@
...
+        let file = files.get(k);
+        console.log(`    file #${k + 1}:`);
+        console.log(`      name: ${file.name().to_str()}`);
+        console.log(`      media_type: ${file.media_type().to_str()}`);
+        console.log(`      src: ${file.src().to_string()}`);
+      }
+    }
+  }
+}
+```
```
rooooooooob, line 47: We can add more documentation later for going form other Metadata but IIRC the bytes API is the one Flint is going to need. I mostly wanted to test out the whole thing directly from JS/wasm.

cip25-wasm/src/lib.rs
```diff
@@ -14,12 +16,10 @@ pub type PolicyIdV1 = String64;
 pub type PolicyIdV2 = Vec<u8>;
```
rooooooooob, line 16: Do we want these to be straight Strings/Byte arrays? Or are these meant to be multiasset policy IDs/asset names that we should try and integrate with the rest of the library? If so then maybe the v1 vs v2 difference could be abstracted away entirely as a serialization detail and we could probably merge the V1 and V2 structs in general in this library. We could also merge `CIP25Metadata` with `LabelMetadata` to make the API simpler. At that point we'd be changing a lot of serialization code though and it'd no longer match up very well with the codegen output.

cip25-wasm/src/lib.rs
```diff
@@ -710,36 +708,40 @@ impl From<MetadataDetails> for core::MetadataDetails {
...
+        core::String64::new(s).map(Self).map_err(Into::into)
     }
 
-    pub fn from_bytes(data: Vec<u8>) -> Result<String64, JsValue> {
-        FromBytes::from_bytes(data).map(Self).map_err(|e| JsValue::from_str(&format!("from_bytes: {}", e)))
+    pub fn to_str(&self) -> String {
+        self.0.to_str().to_owned()
+    }
+
+    pub fn to_bytes(&self) -> Vec<u8> {
```
rooooooooob, line 727: Do we even want to be providing these to/from bytes for all parts of this spec? I don't know why you would want to do it for anything but the root `CIP25Metadata` type.

cip25-wasm/src/lib.rs
SebastienGllmt: I would keep these matching the CDDL as-is. Anything we need to wrap to create a better API should be done as a separate utility type in a separate file since the goal is to never modify the codegen files

SebastienGllmt (approved):

cip25-wasm/src/lib.rs
rooooooooob: For simply wrapping that's fine, but if we want to integrate the policy IDs/asset names with the other types (are they the same right?) we could *massively* simplify the cip25 crate doing that. It would replace ~12 types with only a couple as we could abstract away the v1 vs v2 difference into just encoding differences and a simple enum at the top level that toggles between text vs byte encoding at lower levels. Speaking from our experience in CSL we never once had to do any complete or even significant regeneration after we've finalized the format as any era difference was only just generating the differences and integrating that with the existing code. Or now that we're generating it's because it's such a massive change (split wasm/rust) that any wrappers would be mostly invalidated anyway. I agree that we should reduce the editing of code generated code and use wrappers when it makes sense, but I don't think it can be avoided here unless we like having awkward APIs with many redundant structs and more end-user overload. There would be a lot of conversions though. We would have had 4 auxiliary data types instead of 1 in CSL if we never edited the code, for example.

If we want to not modify it at all I think the best interface we could make would be that there's one single interface on `CIP25Metadata` that is like `get(policyid, assetname) -> Option<details>` (and 1 setter for creating) and then make every other struct in the crate private to hide all those redundant types that should only matter for encoding. Thoughts? Providing a better interface for exploring the full structure is awkward unless we unify the v1 vs v2 policyids/assetnames which unless there's general editing of serialization code to ensure that they're encoded correctly, or would allow you to  end up with a format that's neither v1 nor v2.

Would users want to, say, find all NFTs with a single policy ID? Or do they care more about policyID x assetname pairs? If it's the latter then the simplified wrapper interface is more reasonable (but would still be going between the chain policyid/assetnames and the two v1/v2 ones local to here if we're not editing serialization code).

gostkin (approved):

PR #165, opened 2023-01-15
Cip25 codegen cleanup

SebastienGllmt:
The code generated from this PR uses https://github.com/dcSpark/cddl-codegen/pull/125#pullrequestreview-1249102604 as well as a modified CIP25 cddl spec (see my change to the specs repo)

With these changes, the diff between the codegen output and the code we're checking in is less than 10 lines of code (excluding comments and changes to the serialization file we made to work around lack of support for non-exact types in cddl-codegen)


cip25/src/lib.rs
```diff
@@ -1,91 +1,86 @@
-use std::io::{BufRead, Write};
```
SebastienGllmt, line 1: The diff on all these files are really hard to read since it looks like cddl-codegen doesn't have a stable order for its output. Probably we should change that so diffs are clean in the future (https://github.com/dcSpark/cddl-codegen/issues/126)

PR #166, opened 2023-01-19
Conditional build example with features

gostkin:
Example how we can omit publishing 2 crates


gostkin:
this was just an example PR, so closing it

PR #172, opened 2023-01-31
Fix for non-cip25 fields breaking deserialization

rooooooooob:
Non-cip25 fields were never modifying `CBORReadLen` which caused a deserialization error afterwards as it thought that we hadn't used all the CBOR fields within the current map/array.

This passes the test added in #171 and also addresses the same issue in other parts of the library.


SebastienGllmt (approved):
LGTM

PR #174, opened 2023-01-31
Flaky random-improve tests now try 3 times

rooooooooob:
Before this change I ran the tests 100 times and had 3 failures (i.e. 1 or more of those 5 tests failed).

This can occasionally be annoying if it causes the CI to report a failure.

After this change each one runs 3 times which should make it extremely unlikely to have a test fail as the same test would have to fail 3 times in a row to fail. Using those stats from above that's a <0.0027% (assuming each test 3% chance - likely much lower as that's the chance for only 1 or more of the 5 tests failing) chance of failure.

### DIFF
--- a/rust/Cargo.lock
+++ b/rust/Cargo.lock
@@ -62,6 +62,7 @@ dependencies = [
  "cryptoxide",
  "digest 0.9.0",
  "ed25519-bip32",
+ "flaky_test",
  "fraction",
  "getrandom 0.2.8",
  "hex",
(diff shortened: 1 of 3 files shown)


SebastienGllmt:
Is this really better than just setting a constant seed for the randomness when testing?

rooooooooob:
I wouldn't say it's much better than a fixed seed, but it was a very minor change (6 lines of code - the other 12 were in the cargo lock) to add and should virtually never fail so I don't think it's worse than a constant seed either.

I guess you could look at it as it tests more inputs (considering seeds) since it's not just fixed to one seed. I guess if someone modified things this would guarantee it at least works 1/3 times regardless of seed rather than simply lucking out on the 1 fixed seed.

SebastienGllmt (approved):

PR #175, opened 2023-02-01
Another cip25 test

SebastienGllmt:
This PR adds another test case that fails on CIP25 parsing. It fails with `DeserializeError { location: Some("MetadataDetails"), failure: DefiniteLenMismatch(3, None) }`


rooooooooob (approved):
LGTM. I fixed it locally and will put up a fix PR.

PR #182, opened 2023-02-08
Enable weakrefs in WASM builds

rooooooooob:
This will allow users to not have to explicitly `free()` every single object from WASM, greatly improving the user experience.

People running very allocation-heavy processe using CML can still use `free()` to have more control over deallocation.

`cip25` was tested with this with a heavy allocation example from the browser (running the example in `cip25/wasm/README.md`) many times. Without it, it would run out of memory after 250,000 itertions consistently. With it, and with timeouts to allow the GC to kick in it was able to make it much higher but tended to eventually run out of memory after 1-2 million. This could be something unavoidable in how the GC works (fragmentation? or maybe 50k chunks we were using were too big so it wasn't getting run until it was too late?).

Further testing/investigation might be warranted, but in general if you are using CML for heavy workloads you might still want to do `free()` on the objects used there.


gostkin:
@rooooooooob is there a case when we might not want to enable weakrefs? e.g. should it be a parameter?

Also is there anything that prevents us from merging this PR except conflicts?

rooooooooob:
It's ready to be merged. I'm not aware of reasons it shouldn't be enabled. It only affects WASM builds, and you should still be able to call `free()` on things manually even if it's enabled. I haven't profiled to see if there is any performance differences though in WASM builds between weakrefs off, weakrefs on and explicit freeing, and weakrefs on and implicit freeing in case there is some difference.

gostkin (changes requested):
let's merge after the conflicts are resolved

PR #183, opened 2023-02-16
Add loose nft parsing

SebastienGllmt:
Not all NFTs properly follow CIP25. Notably, some NFTs
- Are missing mandatory keys
- Use different names for the fields for some reason

To tackle this, I created loose NFT parsing logic so that we can use it in Flint to try and detect more NFTs properly

### DIFF
--- a/cip25/rust/src/utils.rs
+++ b/cip25/rust/src/utils.rs
@@ -127,20 +127,17 @@ impl MiniMetadataDetails {
     }
 
     /// loose parsing of CIP25 metadata to allow for common exceptions to the format
+    /// `metadatum` should represent the data where the `MetadataDetails` is in the cip25 structure
     /// TODO: this is not an ideal solution
     ///       ideally: we would have a function that takes in a policy ID
     ///       and would have a lookup map to know which lambda to call to get the name & image depending on the policy ID
     ///       with a fallback to the standard CIP25 definition
     ///       however, since this is a lot of work, we use this temporary solution instead
-    pub fn loose_parse(data: Vec<u8>) -> Result<Self, DeserializeError> {
-        let metadatum = TransactionMetadatum::from_bytes(
-            data,
-        )?;
+    pub fn loose_parse(metadatum: TransactionMetadatum) -> Result<Self, DeserializeError> {
         match metadatum {
             TransactionMetadatum::Map { entries, .. } => {
                 let name: Option<String64> = entries
                 .get(&TransactionMetadatum::new_text("name".to_owned()))
-                .or_else(|| entries.get(&TransactionMetadatum::new_text("name".to_owned())))
                 // for some reason, 1% of NFTs seem to use the wrong case
                 .or_else(|| entries.get(&TransactionMetadatum::new_text("Name".to_owned())))
                 // for some reason, 0.5% of NFTs use "title" instead of name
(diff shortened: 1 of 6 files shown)


cip25/rust/src/utils.rs
```diff
@@ -109,6 +109,87 @@ impl From<&ChunkableString> for String {
...
+    ///       however, since this is a lot of work, we use this temporary solution instead
+    pub fn loose_parse(data: Vec<u8>) -> Result<Self, DeserializeError> {
+        let metadatum = TransactionMetadatum::from_bytes(
+            data,
+        )?;
+        match metadatum {
+            TransactionMetadatum::Map { entries, .. } => {
+                let name: Option<String64> = entries
+                .get(&TransactionMetadatum::new_text("name".to_owned()))
+                .or_else(|| entries.get(&TransactionMetadatum::new_text("name".to_owned())))
```
SebastienGllmt: This is a duplicate (`name` a second time)

cip25/rust/src/utils.rs
```diff
@@ -109,6 +109,87 @@ impl From<&ChunkableString> for String {
...
+    /// TODO: this is not an ideal solution
+    ///       ideally: we would have a function that takes in a policy ID
+    ///       and would have a lookup map to know which lambda to call to get the name & image depending on the policy ID
+    ///       with a fallback to the standard CIP25 definition
+    ///       however, since this is a lot of work, we use this temporary solution instead
+    pub fn loose_parse(data: Vec<u8>) -> Result<Self, DeserializeError> {
+        let metadatum = TransactionMetadatum::from_bytes(
+            data,
+        )?;
+        match metadatum {
```
SebastienGllmt, line 137: The reason this code looks ugly is that we don't have utility functions for MetadatumMap available in the new bindings. This is because the `TransactionMetadatum` type inlines the enum types instead of making them @newtype so we can't target them in a utils function to add more functionality.

cip25/rust/src/utils.rs
```diff
@@ -109,6 +109,87 @@ impl From<&ChunkableString> for String {
...
+                            Err(_) => None,
+                        },
+                        TransactionMetadatum::List { elements, .. } => {
+                            (|| {
+                                let mut chunks: Vec<String64> = vec![];
+                                for i in 0..elements.len() {
+                                    match elements.get(i) {
+                                        Some(TransactionMetadatum::Text { text, .. }) => {
+                                            match String64::new_str(&text) {
+                                                Ok(str64) => chunks.push(str64),
```
rooooooooob, line 169: I guess this is inlined here since this chunking logic is only available in `ChunkableString` when going from CBOR bytes and not a `TransactionMetadatum`? (and converting would be slow) I guess that's okay here and we could move it if it ever becomes useful elsewhere.

cip25/wasm/README.md
```diff
@@ -44,4 +44,11 @@ for (var i = 0; i < policies.len(); ++i) {
     }
   }
 }
+```
+
+We also support loose NFT parsing to try and parse key information out of potentially incorrectly formatted CIP25
```
rooooooooob, line 49: Potentially incorrect? There's an explicit spec so it's undoubtedly incorrectly formatted to be CIP25.

cip25/rust/src/utils.rs
```diff
@@ -109,6 +109,87 @@ impl From<&ChunkableString> for String {
...
+        }
+    }
+
+    /// loose parsing of CIP25 metadata to allow for common exceptions to the format
+    /// TODO: this is not an ideal solution
+    ///       ideally: we would have a function that takes in a policy ID
+    ///       and would have a lookup map to know which lambda to call to get the name & image depending on the policy ID
+    ///       with a fallback to the standard CIP25 definition
+    ///       however, since this is a lot of work, we use this temporary solution instead
+    pub fn loose_parse(data: Vec<u8>) -> Result<Self, DeserializeError> {
```
rooooooooob: This API might be confusing as it's rather opaque. The bytes it's taking in are in particular for the metadata details. Of course, we can't take in one as it would have failed to parse earlier, and taking in a `TransactionMetadatum` is only marginally better (since we're just parsing those bytes anyway into one). I would comment the function to say exactly which part of the cip25 spec this is parsing to avoid confusion.

cip25/rust/src/utils.rs
rooooooooob: And will be removed before the PR is done, right?

rooooooooob (approved):
Overall LGTM code-wise, although I would update the parsing doc comment to say exactly what kind of bytes it's taking in to reduce confusion (+that duplicate you pointed out).

On a non-code level, are we worried this might legitimize invalid NFT formats? They are ~1% of NFTs, but are there any important NFT projects that made these errors or was it mostly people playing around at the start of CIP25? Or do we plan to use this in a way that lets us show the NFTs in Flint but not acknowledging that they're cip25? (I'm not sure how it's displayed/used there nor if there's any way for Flint/etc to use cip25 NFTs in any way that isn't juts displaying them e.g. interaction with dApps).

cip25/wasm/README.md
SebastienGllmt: Potentially incorrect in that valid CIP25 will also be accepted by this function

cip25/rust/src/utils.rs
SebastienGllmt: I feel like maybe `ChunkableString ` and similar types should be provided as utility types in some common crate since this affects all metadata definitions for Cardano (same as String64, etc)
For now, I'll just inline this here

PR #187, opened 2023-02-24
Cml regen new module format

rooooooooob:
CML regen with new module format

also includes direct use of crypto types through cddl-codegen's
integration with `RawBytesEncoding` (https://github.com/dcSpark/cddl-codegen/pull/148)

`address.rs` was manually updated. other files had to have their imports
manually fixed as everything is in a different location vs how CML
outputs things for single-crate output.

includes updates to the spec for style changes + instructions for
generation


specs/babbage/crypto.cddl
```diff
@@ -31,6 +26,43 @@ pool_metadata_hash    = $hash32
...
 ;   "\x02" for Plutus V2 scripts
-scripthash            = $hash28
+;scripthash            = $hash28
+
+;datum_hash = $hash32
+
+KES_signature /= bytes .size 32
+signkeyKES = bytes .size 16
+
+VRF_cert /= [bytes, bytes .size 10]
```
rooooooooob: these are definitely not the right sizes. I'll update this later using the old CML/CSL as a reference as those should have the right sizes.

core/rust/src/serialization.rs
```diff
@@ -195,8 +195,9 @@ pub trait Serialize {
     /// Bytes of a structure using the CBOR bytes as per the CDDL spec
     /// which for foo = bytes will include the CBOR bytes type/len, etc.
     /// This gives the original bytes in the case where this was created
-    /// from bytes originally
-    fn to_original_cbor_bytes(&self) -> Vec<u8> {
+    /// from bytes originally, or will use whatever the specific encoding
+    /// details are present in any encoding details struct for the type.
+    fn to_cbor_bytes(&self) -> Vec<u8> {
```
rooooooooob, line 200: I changed this since:
1) it's what cddl-codegen is outputting
2) the "original" part might be confusing since it's only "original" if you're round-tripping. It also makes it seem like the default one people should use, unless they specifically know they need canonical bytes.

specs/babbage/lib.cddl
```diff
@@ -29,60 +31,61 @@ update = [ proposed_protocol_parameter_updates
...
 mint = { * policy_id => { * asset_name => int64 } }
 
-int64 = -9223372036854775808 .. 9223372036854775807
+int64 = -9223372036854775808 .. 9223372036854775807 ; @no_alias
 
-network_id = 0 / 1
+; not actually 0-256 but codegen doesn't support 0-16
+network_id = uint .size 1 ; @newtype
+; can apparently be more?
+; network_id = 0 / 1
```
SebastienGllmt, line 90: I think we can remove these two comment lines

chain/rust/src/crypto/utils.rs
```diff
@@ -0,0 +1,31 @@
+//impl BootstrapWitness {
```
SebastienGllmt, line 1: Is this meant to all be commented out?

rooooooooob: We haven't migrated the Byron code over yet which this depends on.

rooooooooob:
@SebastienGllmt do we want to handle any of the following in this PR or in separate ones after? (leaving cip36 not compiling in the meantime):

* implement the hand-written structs (plutus const data, bigint, bounded bytes)
* Update CIP36 (since it pulled in some stuff via `ChainCrypto` and could also benefit from this raw bytes encoding stuff + separate all the hand-written stuff into a `utils.rs` like with cip25/the rest of CML now)

If not, I can change those incorrect crypto sizes and mark it as ready for review.

SebastienGllmt:
I have no issue with having it a separate PR

crypto/rust/src/lib.rs
```diff
@@ -615,6 +615,8 @@ impl_hash_type!(VRFVkey, 32);
 impl_hash_type!(KESVkey, 32);
 // same for this signature (but lots of traits aren't implemented for [u8; 448] so we can't)
 //impl_hash_type!(KESSignature, 448);
+// this might actually be one of the other hashes
```
rooooooooob: Does anyone know if this is some other existing hash (block hash, etc) or its own thing? We don't use any existing hash in the old CML code (but the Nonce object has been manually edited there).

specs/babbage/lib.cddl
rooooooooob: What about other places where I have the original code commented out? (e.g. where we use `_CDDL_CODEGEN_EXTERN_TYPE_` / `_CDDL_CODEGEN_RAW_BYTES_TYPE_`)

SebastienGllmt: I think those are fine to keep

crypto/rust/src/lib.rs
SebastienGllmt: It looks like it's its own thing

Notably, in the codebase, the nonce hash is generated by a function
```haskell
mkNonceFromNumber :: Word64 -> Nonce
mkNonceFromNumber =
  Nonce
    . (castHash :: Hash Blake2b_256 Word64 -> Hash Blake2b_256 Nonce)
    . hashWith (BSL.toStrict . B.runPut . B.putWord64be)
```

which seems to be called with the slot number as inputs (So basically a hash of the binary representation of a slot number)

specs/babbage/lib.cddl
rooooooooob: Are there any other changes that need to be done in this PR or is it just that change? Or has it not been fully reviewed yet since it was  draft before?

SebastienGllmt (approved):

specs/babbage/lib.cddl
SebastienGllmt: Nothing else. I just didn't approve since it was still in draft

PR #191, opened 2023-03-08
CIP36 use RawBytesEncoding directly + refactoring

rooooooooob:
Refactor CIP36 to use RawBytesEncoding directly e.g. #187

Refactored CIP36 to have hand-written code be in utils.rs like with CIP25

cip36.cddl updated to correctly follow the basic group clarification + have more things properly named.

### DIFF
--- a/cip36/rust/src/error.rs
+++ b/cip36/rust/src/error.rs
@@ -1,0 +1,10 @@
+#[derive(Debug, thiserror::Error)]
+pub enum CIP36Error {
+    #[error("Empty delegation array")]
+    EmptyDelegationArray,
+    // TODO: can we check this somehow against anything? I don't believe so, so maybe remove this
+    // #[error("Reward wrong network")]
+    // RewardWrongNetwork,
+    #[error("Invalid delegation weights")]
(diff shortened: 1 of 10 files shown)


cip36/rust/src/error.rs
```diff
@@ -0,0 +1,10 @@
+#[derive(Debug, thiserror::Error)]
+pub enum CIP36Error {
+    #[error("Empty delegation array")]
+    EmptyDelegationArray,
+    // TODO: can we check this somehow against anything? I don't believe so, so maybe remove this
```
SebastienGllmt, line 5: TODO?

SebastienGllmt (approved):

cip36/rust/src/error.rs
rooooooooob: This TODO is from what was already in `develop` in `lib.rs`. I can remove it if you want. In the CIP36 spec the stake address must be discriminated with the same network that the tx that contains the metadata is submitted it is on, but then I don't think there's a way to check this in this self-contained cip36 metadata library. It would have to involve extra data either passed in or done after the fact.

PR #192, opened 2023-03-10
BigInt + bounded bytes + ConstrPlutusData

rooooooooob:
#  BigInt + bounded bytes

bounded-bytes logic including adapting for preserving encodings.
This is 100% an encoding detail as we can express the chunking logic in
the encoding variable and expose just a simple bytes type.

bigint hand-written struct to cover the specific big integer encoding
(which can be `Int` too), and also is chunked i.e. bounded bytes

#  ConstrPlutusData

written by hand to avoid having hundreds of variants

### DIFF
--- a/specs/babbage/plutus.cddl
+++ b/specs/babbage/plutus.cddl
@@ -22,7 +22,18 @@ plutus_data =
 
 big_int = _CDDL_CODEGEN_EXTERN_TYPE_
 
-constr_plutus_data = #6.102([constructor: uint, fields: [* plutus_data]])
+; original definition not used to avoid hundreds of variants:
+; constr<a> =
+;    #6.121([* a])
+;  / #6.122([* a])
+;  / #6.123([* a])
+;  / #6.124([* a])
+;  / #6.125([* a])
+;  / #6.126([* a])
+;  / #6.127([* a])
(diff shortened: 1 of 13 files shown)


specs/babbage/plutus.cddl
```diff
@@ -1,3 +1,15 @@
+; bounded_bytes = bytes .size (0..64)
```
SebastienGllmt, line 1: Wouldn't this affect CIP25 as well if the CDDL isn't accurate?

rooooooooob: CIP25 is a subset of the metadata. The way this is defined here is basically the same as how CBOR **indefinite** byte strings are encoded, but with the restriction that *each definite segment* is at most 64-bytes long, whereas how cip25 is defined the bytes there can be encoded indefinite but with *the sum of all definite segments* being 64-bytes max, thus still abiding by this true bounded_bytes definition without needing to specify it.

rooooooooob: actually that's not even an issue with cip25 since bounded bytes it only used in plutus datums, not metadatums so the two things never even touch each other

SebastienGllmt (approved):

PR #194, opened 2023-03-17
Regen byron

rooooooooob:
using specs/byron_minimal.cddl

byron/utils.rs had to be modified a fair bit to deal with changes in
cddl-codegen since then, .cbor bytes support, etc.


rooooooooob:
to see which changes were made from the old code, ignore the first commit in the diff tab. the first commit is just copying over the old byron code directly.

chain/rust/src/byron/utils.rs
```diff
@@ -0,0 +1,496 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum ByronAddressError {
+    #[error("UnknownNetwork: {0}")]
+    UnknownNetwork(ProtocolMagic),
+    #[error("InvalidCRC: found {found}, expected {expected}")]
+    InvalidCRC{ found: Crc32, expected: Crc32 },
+}
+
+impl_hash_type!(AddressId, 28);
+// not sure if this is a hash but it likely is and has the same byte format
```
SebastienGllmt, line 30: AFAIK it was meant to be a hash and then this feature was never added to mainnet in the end

chain/rust/src/byron/utils.rs
```diff
@@ -0,0 +1,496 @@
...
+    // bootstrap era + no hdpayload address
+    pub fn new_simple(xpub: Bip32PublicKey, protocol_magic: Option<ProtocolMagic>) -> Self {
+        let attributes = AddrAttributes::new_bootstrap_era(None, protocol_magic);
+        let addr_type = ByronAddrType::PublicKey;
+        let spending_data = SpendingData::new_spending_data_pub_key(xpub);
+
+        AddressContent::hash_and_create(addr_type, &spending_data, attributes)
+    }
+
+    /// Do we want to remove this or keep it for people who were using old Byron code?
```
SebastienGllmt, line 121: I'm okay with either. If we don't need it, it's not hard to replace imo

specs/byron_minimal.cddl
```diff
@@ -0,0 +1,60 @@
...
+; blake2b224 of sha256 of cbor(byron_addr_type, spending_data, addr_attributes)
+address_id = _CDDL_CODEGEN_RAW_BYTES_TYPE_ ; blake2b224 
+stakeholder_id = _CDDL_CODEGEN_RAW_BYTES_TYPE_; blake2b224
+crc32 = _CDDL_CODEGEN_EXTERN_TYPE_ ; uint .size 4 ; @newtype
+protocol_magic = _CDDL_CODEGEN_EXTERN_TYPE_ ; uint .size 4 ; @newtype
+hd_address_payload = bytes ; @newtype
+
+; Addresses
+
+; cddl had bootstrap as (1, uint) but byron/mod.rs had no uint field.
```
SebastienGllmt, line 22: don't remember off the top of my head the history to this, but I can look into it to try and remember

chain/rust/src/byron/utils.rs
rooooooooob: alright I'll just leave it then. whatever it was meant to be, the hash type doesn't have any explicitly "hash" things I guess. it's just a fixed size byte buffer that can be converted to hex/etc.

chain/rust/src/byron/utils.rs
rooooooooob: After having run into some type annotations needed in a few places when trying to use the From/Into for addresses, I think we could just keep the to_address()/from_address() things everywhere since it'd also be consistent with wasm (where we can't expose traits) and would help people migrating from CSL/old CML.

specs/byron_minimal.cddl
```diff
@@ -0,0 +1,60 @@
...
+stake_distribution = [
+  ; @name SingleKey
+  tag: 0, stakeholder_id //
+  ; @name BootstrapEra
+  bootstrap_era: 1
+]
```
rooooooooob, line 28: I had to put the field name as `bootstrap_era` too since cddl-codegen's dcSpark/cddl-codegen#153 ignores `@name` for single-field group choices. i.e. `tag: 1` as the field would make the entire variant come out as `StakeDistribution::Tag` instead of just the fixed field during serialization.

specs/byron_minimal.cddl
SebastienGllmt: I think this may have just been a mistake on my part. The original cddl is `addrdistr = [1] / [0, stakeholderid]` and I think maybe I added a `uint` for the `[1]` case even though it shouldn't have been there

SebastienGllmt (approved):
LGTM

PR #195, opened 2023-03-22
Value + MultiAsset(/Mint) custom code

rooooooooob:
Based on old CML's Value / MultiAsset classes.

AssetBundle generalizes MultiAsset/Mint to work on any integer-like value type.

Includes both checked and clamp operations, with the clamping/checking applied uniformly to both the AssetBundle and Coin for Value, a difference from the old Value where both used clamping operation for the MultiAsset part.

Value was very lightly hand-edited to match closer the old format, although without the multiasset part being an Option. This required one extra field for ValueEncoding to differentiate between a coin-only Value and one using the multiasset format but without any assets (e.g. empty map).

Only includes the rust API. Once this is finalized we can work on the wasm API.


chain/rust/src/assets.rs
```diff
@@ -0,0 +1,595 @@
...
+}
+
+impl<T: num::CheckedSub + num::Bounded + num::Zero + Ord> ClampedSub for T {
+    fn clamped_sub(&self, rhs: &Self) -> Self {
+        self.checked_sub(rhs).unwrap_or_else(|| if *rhs < T::zero() { T::max_value() } else { T::min_value() })
+    }
+}
+
+impl<T> ClampedSub for AssetBundle<T>
+where T: num::CheckedAdd + num::CheckedSub + num::Zero + num::Bounded + Copy + Clone + ClampedSub {
```
rooooooooob, line 54: How do we feel about traits vs methods for consistency between rust/wasm vs idiomatic rust? Not just here, but in general.

chain/rust/src/assets.rs
```diff
@@ -0,0 +1,595 @@
...
+impl<T> AssetBundle<T>
+    where T: num::CheckedAdd + num::CheckedSub + num::Zero + num::Bounded + Copy + Clone,
+{
+    /// Set the value of policy_id:asset_name to value.
+    /// Returns the previous value, or None if it didn't exist
+    pub fn set(&mut self, policy_id: PolicyId, asset_name: AssetName, value: T) -> Option<T> {
+        self.0.entry(policy_id).or_default().insert(asset_name, value)
+    }
+
+    /// Get the value of policy_id:asset_name. If it doesn't exist, 0 is returned.
```
rooooooooob: Is it more useful for users to have this turn into 0 if it's not present, or have an Option<T> returned?

chain/rust/src/assets.rs
```diff
@@ -0,0 +1,595 @@
...
+        }
+        Some(Self(bundle))
+    }
+}
+
+pub type Mint = AssetBundle<i64>;
+
+pub type MultiAsset = AssetBundle<u64>;
+
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
```
rooooooooob, line 204: One other possibility, if we really want to maximize direct code generation, is to instead modify this API over top of 100% directly generated `value = coin / [coin, multiasset]`. That would result in something like:
```rust
pub enum Value {
    Coin {
        coin: Coin,
        #[serde(skip)]
        coin_encoding: Option<Sz>,
    },
    Multiasset(MultiAssetValue),
}

pub struct MultiAssetValue {
    coin: Coin,
    multiasset: MultiAsset,
    #[serde(skip)]
    encoding: Option<MultiAssetValueEncoding>,
}
```

I left it how it is (we generated this originally off of code like `value = [coin, multiasset]` then I just added 1 variable to `ValueEncoding` + spliced in the `Coin` part of (de)serialization) to closer match old CML's `Value` class and to simplify the multiasset handling for such a core class like `Value`.  Users can check `Value::has_multiassets()` if they want to check.

chain/rust/src/assets.rs
SebastienGllmt: I feel like a lot of the point of this project was to allow idiomatic rust in the Rust file, so I would prefer idiomatic Rust instead of fine-tuning things for the WASM wrapper

chain/rust/src/assets.rs
rooooooooob: i.e. then we'd have code like `assets/mod.rs` / `assets/cbor_encodings.rs` / `assets/serialization.rs` containing that code and then all this other stuff in `assets/utils.rs`.

chain/rust/src/assets.rs
SebastienGllmt: I prefer Option. I feel like returning zero will cause at least one person to get it wrong

chain/rust/src/transaction/serialization.rs
```diff
@@ -2649,7 +2649,7 @@ impl Deserialize for TransactionBody {
                 update,
                 auxiliary_data_hash,
                 validity_interval_start,
-                mint,
+                mint: mint.map(Into::into),
```
rooooooooob, line 2652: This is the one area we have to modify serialization code in exchange for being able to have an `AssetBundle` class. It's trivial to do so I don't think it's a big deal, and having our class like that (with Deref traits) lets you both treat it like it's still an OrderedHashMap in most contexts and also implement our own methods/traits over top of it.

core/rust/src/ordered_hash_map.rs
```diff
@@ -5,6 +5,13 @@ pub struct OrderedHashMap<K, V>(linked_hash_map::LinkedHashMap<K, V>)
 where
     K: Hash + Eq + Ord;
 
+impl<K, V> Default for OrderedHashMap<K, V>
```
rooooooooob, line 8: I didn't derive `Default` as `ordered_hash_map::OrderedHashMap` only implements it when your generic parameters implement certain traits. We don't use a non-default hasher though so `new()` is fine for us.

chain/rust/src/assets.rs
rooooooooob: What if we provide both and have the zero one like `get_or_zero()`? (or whatever name would be better)

rooooooooob: or `count()`, `asset_count()`, etc.

SebastienGllmt: What would the benefit of `get_or_zero` be over just using an `unwrap_or` ?

counts might be useful to calculate the min utxo related to this bundle, but beyond that I'm not sure how many utility functions we want to expose given you can also just deref the type to get the underlying map and call utility functions on that, right?

rooooooooob: Yeah it would be more for convenience + there would be no `unwrap_or` in WASM and having it only in the WASM wrapper might be weird. We can leave it as just `get() -> Option<T>` if we want to limit the utility functionality though. For Cardano semantics is there a reason you'd ever have an explicit 0 instead of not having it in a multiasset/mint, and if you did would it matter (besides CBOR serialization details)? I guess for wallets/explorers/etc you might want to show it even if it's 0 maybe for the tx details, although you'd probably be iterating the object there instead of checking a specific policy id / asset name combo.

SebastienGllmt: I believe the only time you're ever allowed a 0 value of anything is when you're withdrawing 0 ADA from a reward account

rooooooooob: @SebastienGllmt Is there anything else here that needs changing before we can merge this? Or does it need more reviewing?

So should we just switch it to returning an `Option<T>`? Or if there cannot be a 0 value for any of these (since these are assets not ADA) keep it how it is or what? I don't have a strong preference. I guess what I said earlier about WASM not having `unwrap_or` isn't true since you can use null-coalescing in JS like `foo.get(id, name) ?? 0` if they need that so I guess the `Option<T>` is fine. The only possible issue is from what I can tell `None` is `undefined`, not `null` so if anyone directly does arithmetic on it you'd get NaN instead of 0 though with `null`, but only when the asset is absent. I guess lots of static analysis/type checking things would find that kind of issue though.

chain/rust/src/assets.rs
```diff
@@ -0,0 +1,595 @@
...
+            .get(policy_id)
+            .and_then(|assets| assets.get(asset_name))
+            .copied()
+            .unwrap_or(T::zero())
+    }
+
+    /// Adds to bundles together, checking value bounds.
+    /// If an overflow would occur, returns None instead.
+    /// Does not modify self, and instead returns the result.
+    pub fn checked_add(&self, rhs: &Self) -> Option<Self> {
```
gostkin, line 126: i think for the calculations it makes sense to add some tests (checked_add, checked_sub, and clamped versions)

chain/rust/src/assets.rs
gostkin: I think enum is not the most convenient solution for the `Value`. E.g. we can end up in a strange situation with `Multiasset` option, where `multiasset` field has no assets

gostkin (approved):
Looks good, I've added an issue for tests coverage https://github.com/dcSpark/cardano-multiplatform-lib/issues/198

chain/rust/src/assets.rs
rooooooooob: Yeah it's more of how far we want to take the idea of only writing custom code _on top of_ the generated code, rather than replacing it. And the only way to generate code directly here would be with the enum since it's a type choice.

chain/rust/src/assets.rs
SebastienGllmt: yeah, returning either `undefined` or `null` in JS is totally fine. Almost everybody now-a-days uses Typescript or eslint which would warn about invalid arithmetic on possible undefined values

PR #197, opened 2023-04-04
Other builders new format

rooooooooob:
Current status:

- [x] Other builders / utility code (rust)
- [x] Update other builder tests

To be done in later PRs
- tx_builder.rs (rust)
- WASM bindings
- Update all other tests


rooooooooob:
Ignore the first commit when reviewing to see the diffs vs the old code. The first commit is just directly copying the old builder code.

chain/rust/src/lib.rs
```diff
@@ -347,6 +349,17 @@ pub struct Value {
     pub encodings: Option<ValueEncoding>,
 }
 
+// TODO: REMOVE after merging value/multiasset PR
```
SebastienGllmt: TODO

chain/rust/src/crypto/utils.rs
```diff
@@ -11,21 +18,23 @@
...
-//         let address_content = AddressContent::new_simple(&Bip32PublicKey(key), protocol_magic);
-//         Ok(address_content)
-//     }
-// }
+    fn try_from(wit: BootstrapWitness) -> Result<Self, Self::Error> {
+        // let protocol_magic = wit.attributes.protocol_magic;
+        // let key = chain_crypto::PublicKey::<chain_crypto::ed25519_derive::Ed25519Bip32>::try_from(wit)?;
+        // let address_content = AddressContent::new_simple(Bip32PublicKey(key), protocol_magic);
+        // Ok(address_content)
+        todo!("redo byron address attributes to not be just a bytestring")
```
SebastienGllmt: TODO

chain/rust/src/crypto/hash.rs
```diff
@@ -0,0 +1,136 @@
...
+        for redeemer in redeemers {
+            redeemer.serialize(&mut buf, true).unwrap();
+        }
+        if let Some(datums) = datums {
+            buf.write_array(cbor_event::Len::Len(datums.len() as u64)).unwrap();
+            for datum in datums {
+                datum.serialize(&mut buf, true).unwrap();
+            }
+        }
+        todo!("language views encoding");
```
SebastienGllmt: TODO

chain/rust/src/builders/witness_builder.rs
```diff
@@ -0,0 +1,698 @@
...
+        }
+
+        Ok(self.clone().build())
+    }
+}
+
+pub fn merge_fake_witness(builder: &mut TransactionWitnessSetBuilder, required_wits: &RequiredWitnessSet) {
+    let mut remaining_wits = required_wits.clone();
+    builder.vkeys.keys().for_each(|key| { remaining_wits.vkeys.remove(&key.hash()); });
+    todo!("need to adjust byron addr attributes to not just be bytes");
```
SebastienGllmt: TODO

chain/rust/src/builders/withdrawal_builder.rs
```diff
@@ -0,0 +1,119 @@
...
+            address: self.address.clone(),
+            amount: self.amount,
+            aggregate_witness: Some(InputAggregateWitnessData::NativeScript(native_script.clone(), witness_info.clone())),
+            required_wits,
+        })
+    }
+
+    pub fn plutus_script(&self, partial_witness: &PartialPlutusWitness, required_signers: &RequiredSigners) -> Result<WithdrawalBuilderResult, WithdrawalBuilderError> {
+        let mut required_wits = RequiredWitnessSet::default();
+        todo!("the line below won't work until we regenerate RequiredSigners with https://github.com/dcSpark/cddl-codegen/issues/164 fixed");
```
SebastienGllmt: TODO

chain/rust/src/builders/withdrawal_builder.rs
```diff
@@ -0,0 +1,119 @@
...
+    address::RewardAddress,
+    certs::StakeCredential,
+    transaction::RequiredSigners,
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum WithdrawalBuilderError {
+    #[error("Missing the following witnesses for the withdrawal: {0:?}. ")]
+    MissingWitnesses(RequiredWitnessSet),
+    //#[error("Withdrawal required a script, not a payment key: {}", .to_address().to_bech32(None))]
```
SebastienGllmt, line 16: duplicate line?

chain/rust/src/builders/utils.rs
```diff
@@ -0,0 +1,126 @@
+//use crate::auxdata::{AuxiliaryData, GeneralTransactionMetadata};
+
+/// this value is large enough to trigger cbor_event to force writing a u64 cbor type
+/// when used inside `write_type`.
+/// This is useful to calculate payloads assuming the max possible size that can be used
+/// since u64 in CBOR are fixed size (will prefix with 0s to stay a constant width regardless of value)
+pub const force_u64: u64 = 0x01_00_00_00_00; 
+
+// impl GeneralTransactionMetadata {
```
SebastienGllmt, line 3: Why is all of this commented out?

chain/rust/src/builders/mod.rs
```diff
@@ -0,0 +1,9 @@
+pub mod certificate_builder;
+pub mod input_builder;
+pub mod redeemer_builder;
+pub mod output_builder;
+//pub mod tx_builder;
```
SebastienGllmt, line 5: is this meant to be commented out?

chain/rust/src/builders/mint_builder.rs
```diff
@@ -0,0 +1,61 @@
...
+            aggregate_witness: Some(InputAggregateWitnessData::NativeScript(native_script.clone(), witness_info.clone())),
+            required_wits,
+        }
+    }
+
+    pub fn plutus_script(&self, partial_witness: &PartialPlutusWitness, required_signers: RequiredSigners) -> MintBuilderResult {
+        let mut required_wits = RequiredWitnessSet::default();
+
+        let script_hash = partial_witness.script.hash();
+        todo!("the line below won't work until we regenerate RequiredSigners with https://github.com/dcSpark/cddl-codegen/issues/164 fixed");
```
SebastienGllmt: TODO

chain/rust/src/builders/input_builder.rs
```diff
@@ -0,0 +1,135 @@
...
+            input: self.input.clone(),
+            utxo_info: self.utxo_info.clone(),
+            aggregate_witness:Some(InputAggregateWitnessData::NativeScript(native_script.clone(), witness_info.clone())),
+            required_wits,
+        })
+    }
+
+    pub fn plutus_script(&self, partial_witness: &PartialPlutusWitness, required_signers: &RequiredSigners, datum: &PlutusData) -> Result<InputBuilderResult, InputBuilderError> {
+        let mut required_wits = RequiredWitnessSet::default();
+        todo!("the line below won't work until we regenerate RequiredSigners with https://github.com/dcSpark/cddl-codegen/issues/164 fixed");
```
SebastienGllmt: TODO

chain/rust/src/builders/certificate_builder.rs
```diff
@@ -0,0 +1,202 @@
...
+        Ok(CertificateBuilderResult {
+            cert: self.cert.clone(),
+            aggregate_witness: if contains { Some(InputAggregateWitnessData::NativeScript(native_script.clone(), witness_info.clone())) } else { None },
+            required_wits,
+        })
+    }
+
+    pub fn plutus_script(&self, partial_witness: &PartialPlutusWitness, required_signers: &RequiredSigners) -> Result<CertificateBuilderResult, CertBuilderError> {
+        let mut required_wits = RequiredWitnessSet::default();
+        todo!("the line below won't work until we regenerate RequiredSigners with https://github.com/dcSpark/cddl-codegen/issues/164 fixed");
```
SebastienGllmt: TODO

chain/rust/src/builders/mint_builder.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::{
+    PolicyId,
+    AssetName,
+    NativeScript,
+    transaction::RequiredSigners,
+};
+
+#[derive(Clone)]
+pub struct MintBuilderResult {
+    pub(crate) policy_id: PolicyId,
```
gostkin: why `pub(crate)`? I think it makes sense to allow access to the fields

chain/rust/src/builders/output_builder.rs
```diff
@@ -0,0 +1,147 @@
...
+        Ok(SingleOutputBuilderResult {
+            output,
+            communication_datum: self.communication_datum.clone()
+        })
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct SingleOutputBuilderResult {
+    output: TransactionOutput,
```
gostkin, line 124: pub?

chain/rust/src/builders/input_builder.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    };
+    if let Address::Byron(byron) = utxo_info.address() {
+        required_witnesses.add_bootstrap(byron);
+    }
+}
+
+#[derive(Clone, Debug)]
+pub struct InputBuilderResult {
+    pub(crate) input: TransactionInput,
```
gostkin: Why pub(crate)? there's no way to access fields externally when you get this result, it's not very convenient

chain/rust/src/builders/certificate_builder.rs
```diff
@@ -0,0 +1,202 @@
...
+        },
+        // no witness as there is no single core node or genesis key that posts the certificate
+        Certificate::MoveInstantaneousRewardsCert(_cert) => {},
+    };
+    Ok(())
+}
+
+#[derive(Clone)]
+pub struct CertificateBuilderResult {
+    pub(crate) cert: Certificate,
```
gostkin: same question about `pub(crate)`

chain/rust/src/min_ada.rs
```diff
@@ -0,0 +1,397 @@
...
+    transaction::{TransactionOutput, DatumOption, ScriptRef, BabbageTxOut},
+};
+
+#[derive(Debug, thiserror::Error)]
+pub enum MinAdaError {
+    #[error("Arithmetic overflow")]
+    ArithmeticOverflow,
+}
+
+// /// Provide backwards compatibility to Alonzo by taking the max min value of both er
```
gostkin, line 18: is this part needed?

chain/rust/src/lib.rs
gostkin: we can merge already I beleive?

chain/rust/src/builders/witness_builder.rs
```diff
@@ -0,0 +1,698 @@
...
+            _ => None
+        }
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub struct RequiredWitnessSet {
+    // note: the real key type for these is Vkey
+    // but cryptographically these should be equivalent and Ed25519KeyHash is more flexible
+    pub(crate) vkeys: BTreeSet<Ed25519KeyHash>,
```
gostkin: What do you think about allowing the access to the fields?

chain/rust/src/builders/withdrawal_builder.rs
```diff
@@ -0,0 +1,119 @@
...
+        StakeCredential::Script { hash, .. } => {
+            required_witnesses.add_script_hash(hash);
+            // recall: no datum hash for reward withdrawals
+        },
+    }
+}
+
+#[derive(Clone)]
+pub struct WithdrawalBuilderResult {
+    pub(crate) address: RewardAddress,
```
gostkin: same question

chain/rust/src/builders/tx_builder.rs
```diff
@@ -0,0 +1,4966 @@
...
+}
+
+fn min_fee_with_exunits(tx_builder: &TransactionBuilder) -> Result<Coin, JsError> {
+    let full_tx = fake_full_tx(tx_builder, tx_builder.build_body()?)?;
+    // we can't know the of scripts yet as they can't be calculated until we build the tx
+    ledger::alonzo::fees::min_fee(&full_tx, &tx_builder.config.fee_algo, &tx_builder.config.ex_unit_prices)
+}
+
+#[wasm_bindgen]
+pub enum CoinSelectionStrategyCIP2 {
```
gostkin, line 89: what do you think if we divide the file to multiple ones: e.g. separate file per struct. it's too big now

chain/rust/src/builders/tx_builder.rs
```diff
@@ -0,0 +1,4966 @@
...
+
+#[wasm_bindgen]
+#[derive(Clone, Debug, Default)]
+pub struct TransactionBuilderConfigBuilder {
+    fee_algo: Option<LinearFee>,
+    pool_deposit: Option<BigNum>,      // protocol parameter
+    key_deposit: Option<BigNum>,       // protocol parameter
+    max_value_size: Option<u32>,       // protocol parameter
+    max_tx_size: Option<u32>,          // protocol parameter
+    /// TODO: remove once Babbage is on mainnet
```
gostkin, line 126: babbage is on mainnet already

chain/rust/src/builders/tx_builder.rs
```diff
@@ -0,0 +1,4966 @@
...
+    ex_unit_prices: ExUnitPrices, // protocol parameter
+    costmdls: Costmdls,           // protocol parameter
+    collateral_percentage: u32,   // protocol parameter
+    max_collateral_inputs: u32,   // protocol parameter
+    prefer_pure_change: bool,
+}
+
+#[wasm_bindgen]
+#[derive(Clone, Debug, Default)]
+pub struct TransactionBuilderConfigBuilder {
```
gostkin, line 120: could we add impls for mainnet / testnets with a note that the params may change? and the descriptions where to get actual parameters. now it's not super friendly for end users (which might have no idea about cardano parameters)

specs/babbage/transaction.cddl
```diff
@@ -41,7 +41,7 @@ alonzo_tx_out = [
 ]
 babbage_tx_out = {
     0 : address,      ; @name address
-    1 : value,        ; @name value
```
gostkin, line 44: why value -> amount?

chain/rust/src/builders/withdrawal_builder.rs
rooooooooob: I forgot to put a comment in this github PR after asking. I wasn't able to figure out how to call functions in this auto-format string in the thiserror macro to mimic the previous error message (bech32). Is it okay to just debug display the address? Or would anyone happen to know how to print it like that? I guess I could always manually implement this error type and its traits instead.

chain/rust/src/builders/mod.rs
rooooooooob: For now, yes. The rest of the builders all compile since they've been updated. The tx_builder.rs does not yet. I will uncomment this once that's ready. It's a draft PR still.

specs/babbage/transaction.cddl
rooooooooob: For consistency, to match the name used in the other 2 variants.

chain/rust/src/builders/withdrawal_builder.rs
SebastienGllmt: not sure. Bech32 is not to have, but hex is better than nothing

chain/rust/src/builders/tx_builder.rs
SebastienGllmt: yeah, this come from the old code that was copy-pasted. No need to change it in this PR

chain/rust/src/builders/tx_builder.rs
SebastienGllmt: The parameters can change over time which is why we hesitated to add defaults to CML

gostkin: they can change, but it's not frequent as far as I see. why not to simplify the setup? (of course saying that the parameters can change and keeping the way to configure them manually)

chain/rust/src/builders/tx_builder.rs
gostkin: @SebastienGllmt , we can do it later, imo makes sense to get rid of outdated functionality

gostkin: issue for that -- https://github.com/dcSpark/cardano-multiplatform-lib/issues/200

chain/rust/src/builders/utils.rs
rooooooooob: Not referred to so far. I'll probably remove it after finishing migrating `tx_builder.rs` if it's not needed. That is unless we still want them. It just seemed like more of the helper functionality to help with WASM making many temporaries like the ones we talked about in the other builders.

chain/rust/src/min_ada.rs
rooooooooob: @SebastienGllmt do we need any previous era functionality that isn't just parsing on-chain block cbor?

chain/rust/src/builders/witness_builder.rs
rooooooooob: Done as discussed.

chain/rust/src/builders/redeemer_builder.rs
```diff
@@ -0,0 +1,330 @@
...
+        if let Some(data) = plutus_data {
+            self.reward.insert(result.address.clone(), Some(UntaggedRedeemerPlaceholder::JustData(data.clone())));
+        } 
+        else {
+            self.reward.insert(result.address.clone(), None);
+        }
+        
+    }
+
+    pub fn add_cert(&mut self, result: &CertificateBuilderResult) {
```
rooooooooob, line 180: I noticed this (as well as `TransactionWitnessSetBuilder`) are different from the other builders API-wise. @SebastienGllmt do you remember if there was a reason for this? They modify and don't return, whereas other ones return self to allow for function chaining. Was there a performance difference on WASM due to all the cloning instead of moving like in rust? Or some other reason?

chain/rust/src/crypto/hash.rs
rooooooooob: Done now.

chain/rust/src/builders/tx_builder.rs
rooooooooob: The vast majority of the file is for tests. ~1500 is the actual code. It might shrink once rewritten for idiomatic rust code. Here's the approximate breakdown:
* add_change_if_needed: ~300 (free-floating pub function that mutates the tx builder)
* SignedTxBuilder: ~40
* TxRedeemerBuilder: ~80

I don't think per-struct makes sense for breaking it up since the non-tx-builder structs are tiny. You'd have to start breaking off methods off of the tx builder struct itself to make it meaningfully smaller. The coin selection functions are ~250 lines for example.

chain/rust/src/builders/tx_builder.rs
rooooooooob: How many people using CML would actually pull from up to date info instead? I wonder how many people would, instead, copy and paste parameters they got somewhere and were up to date when they made it but then go out of date later. I guess it shifts responsibility from us to the users.

gostkin: > How many people using CML would actually pull from up to date info instead? I wonder how many people would, instead, copy and paste parameters they got somewhere and were up to date when they made it but then go out of date later. I guess it shifts responsibility from us to the users.

if we explain somewhere in the docs how to get actual parameters / provide tool for that, imo would be great as well from ux perspective

gostkin: e.g. at least give a link to https://cexplorer.io/params in the docs

gostkin:
btw @rooooooooob can we make `cargo build --release` compiling for the root project? I see we're missing `chain/genesis` for that

gostkin:
> btw @rooooooooob can we make `cargo build --release` compiling for the root project? I see we're missing `chain/genesis` for that

If we resolve this we can enable a compile check at least

rooooooooob:
>btw @rooooooooob can we make cargo build --release compiling for the root project? I see we're missing chain/genesis for that

Is that in scope here? We didn't touch any of those files. I could do another PR that updates non-builder tests to compile + gets that genesis to compile on release so we can get the compile + tests (if it's not too much work) checks working there.

chain/rust/src/builders/tx_builder.rs
rooooooooob: Looking at that graph there's some periods of time where the parameters were changing more often than I had expected. I think I'm more on the side of not providing it now, even if there are a lot of periods where it's pretty stable. Improving the docs about it is a great idea though.

gostkin:
> > btw @rooooooooob can we make cargo build --release compiling for the root project? I see we're missing chain/genesis for that
> 
> Is that in scope here? We didn't touch any of those files. I could do another PR that updates non-builder tests to compile + gets that genesis to compile on release so we can get the compile + tests (if it's not too much work) checks working there.

no, that's not in the scope of builders pr

gostkin (approved):

chain/rust/src/builders/redeemer_builder.rs
SebastienGllmt: I don't recall any deep reason. These were originally meant to be Rust-only I think but I haven't double-checked how they will be used going forward

PR #210, opened 2023-05-03
TxBuilder refactor to split format

### LINKED ISSUE #202: TxBuilder refactoring
Refactor tx_builder.rs (rust)

rooooooooob:
Closes #202 

Contains only rust crate updates

~Tests are commented out for now as there are thousands and thousands  of lines of them. I will update them later.~

~Tests are updated, but only 67/94 pass. We will need to look into them all one by one vs the originals.~

Tests are all now passing. Several non-test issues were fixed by this as well.


chain/rust/src/assets.rs
```diff
@@ -26,15 +26,45 @@ pub enum AssetArithmeticError {
 }
 
 /// Bundle of assets within range of T, grouped by PolicyID then AssetName
-#[derive(Clone, Debug, Default, PartialEq, Hash, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+#[derive(Clone, Default, PartialEq, Hash, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
 pub struct AssetBundle<T>(OrderedHashMap<PolicyId, OrderedHashMap<AssetName, T>>);
 
+impl<T: std::fmt::Debug> std::fmt::Debug for AssetBundle<T> {
```
rooooooooob, line 32: I'm not sure if we want this to be the default `Default` implementation or instead a separate method since it can truncate data (the middle parts of policy ids and possibly asset names). It was definitely far more readable when debugging these tests vs the default implementation though.

chain/rust/src/builders/utils.rs
```diff
@@ -99,22 +99,24 @@
 //     }
 // }
 
-#[cfg(test)]
-mod tests {
-    use super::*;
+// #[cfg(test)]
```
gostkin: why commented?

chain/rust/src/byron/mod.rs
```diff
@@ -191,16 +192,17 @@ impl StakeDistribution {
...
-    #[test]
-    fn tx_output_decoding() {
-        let tx_out = ByronTxout::from_bytes(
-            hex::decode("8282d818582183581cc6eb29e2cbb7b616b28c83da505a08253c33ec371319261ad93e558ca0001a1102942c1b00000005f817ddfc").unwrap()
-        ).unwrap();
-        assert_eq!(tx_out.address().to_base58(), "Ae2tdPwUPEZGexC4LXgsr1BJ1PppXk71zpuRkboFopVpSDcykQvpyYJXCJf");
-        assert!(tx_out.to_json().unwrap().contains("Ae2tdPwUPEZGexC4LXgsr1BJ1PppXk71zpuRkboFopVpSDcykQvpyYJXCJf"));
-    }
-}
+// #[cfg(test)]
```
gostkin, line 195: why commented?

chain/rust/src/fees.rs
```diff
@@ -0,0 +1,639 @@
...
+    // TODO: the fee should be 0 if all inputs are genesis redeem addresses
+    min_no_script_fee(tx, linear_fee)?
+        .checked_add(min_script_fee(tx, ex_unit_prices)?)
+        .ok_or(ArithmeticError::IntegerOverflow)
+}
+
+// Note: all the tests below are disabled
+// This is because these tests are meant to run on Alonzo-era tx formats to give the right result
+// But currently CML doesn't allow you to encode txs in old formats explicitly
+// which means these tests return the wrong value until we add this feature
```
gostkin: do we plan to have this feature?

chain/rust/src/builders/tx_builder.rs
```diff
@@ -97,181 +183,141 @@ pub enum CoinSelectionStrategyCIP2 {
...
+            key_deposit: self.key_deposit.ok_or_else(|| TxBuilderError::UninitializedField(TxBuilderConfigField::KeyDeposit))?,
+            max_value_size: self.max_value_size.ok_or_else(|| TxBuilderError::UninitializedField(TxBuilderConfigField::MaxValueSize))?,
+            max_tx_size: self.max_tx_size.ok_or_else(|| TxBuilderError::UninitializedField(TxBuilderConfigField::MaxTxSize))?,
+            coins_per_utxo_byte: self.coins_per_utxo_byte.ok_or_else(|| TxBuilderError::UninitializedField(TxBuilderConfigField::CoinsPerUtxoBytes))?,
+            ex_unit_prices: self
                 .ex_unit_prices
-                .ok_or_else(|| JsError::from_str("uninitialized field: ex_unit_prices"))?,
-            costmdls: if cfg.costmdls.is_some() {
-                cfg.costmdls.unwrap()
+                .ok_or_else(|| TxBuilderError::UninitializedField(TxBuilderConfigField::ExUnitPrices))?,
```
gostkin, line 289: why collateral and ex units are always required for tx builder config? not all the transactions need these params. besides that, do you know why we count fee with ex units only while balancing the change, why it is not done e.g. while selecting inputs?

chain/rust/src/builders/tx_builder.rs
```diff
@@ -559,197 +601,191 @@ impl TransactionBuilder {
...
-            let mut new_assets = result.assets.clone();
-            old_assets.0.append(&mut new_assets.0);
-            old_assets
-        };
-        mint.insert(&result.policy_id, &assets);
+    pub fn add_mint(&mut self, result: MintBuilderResult) {
+        self.witness_builders.redeemer_set_builder.add_mint(&result);
+        self.witness_builders.witness_set_builder.add_required_wits(result.required_wits.clone());
+        // TODO: old behavior was to overwrite if for some reason there was already an asset here.
+        // Should we keep that or should we return an error or should we sum them?
```
gostkin: return error is the safest

chain/rust/src/builders/tx_builder.rs
```diff
@@ -559,197 +601,191 @@ impl TransactionBuilder {
...
     }
 
     pub fn set_validity_start_interval(&mut self, validity_start_interval: &Slot) {
         self.validity_start_interval = Some(*validity_start_interval)
     }
 
-    pub fn get_certs(&self) -> Option<Certificates> {
-        self.certs.clone()
-    }
+    // TODO: remove?
```
gostkin: looks like it's not used, so yes

chain/rust/src/builders/tx_builder.rs
```diff
@@ -767,68 +803,66 @@ impl TransactionBuilder {
...
 
         Ok(())
     }
 
-    pub fn collateral(&self) -> Option<TransactionInputs> {
-        self.collateral
-            .as_ref()
-            .map(|list| TransactionInputs(list.iter().map(|input| input.input.clone()).collect()))
-    }
+    // TODO: remove?
```
gostkin: 👍🏻

gostkin: it's more like general question : whether we want to allow users to access tx builder fields or not. so we either allow all or restrict all

gostkin (commented):
overall looks good

chain/rust/src/byron/mod.rs
rooooooooob: Test doesn't compile and is unrelated to the builders. I just commented it out so I could get `cargo test` running on all tests here. I will get it working in the next PR as it's not builder related.

chain/rust/src/builders/utils.rs
rooooooooob: I think I shall just remove this entire file instead. The functionality here was just done form scratch inside of the metadata utils instead of trying to adapt what was here to all the differences between auxdata struct now vs then. It's not really 100% builder either so it doesn't make sense to be here.

chain/rust/src/fees.rs
rooooooooob: They were commented out in the old CML as well, I guess from when we updated to babbage. I got rid of the legacy fee calculation/etc when refactoring now. We had talked about this previously and IIRC decided that we want to support parsing of every era, but for creation/builders it makes more sense to only support the current era. I think I will just remove these commented out tests then.

chain/rust/src/builders/tx_builder.rs
rooooooooob: I just adapted how the code was from the old builders. I assume we just made it mandatory since you'd probably have all the protocol parameters available. We could probably refactor the builders to throw this error later on e.g. if you tried making a tx that used plutus stuff but didn't initialize it so you don't have to provide it always. I'm not sure about the ex units when selecting inputs. What would be needing it for input selection fee-wise that we could reasonably know? We treat plutus as a black box of bytes since we don't have access to an interpreter/etc. It shouldn't affect CBOR sizes since AFAIK the only place that uses it CBOR-wise would be as part of a hash so the size wouldn't change regardless of what the exunits are. I could be misunderstanding or overlooking something though. I just based these changes on the existing code we had in the old crate.

chain/rust/src/builders/tx_builder.rs
rooooooooob: I'm not sure if we want to or not. I suppose we could remove these then if anyone asks we could add them.

chain/rust/src/builders/tx_builder.rs
rooooooooob: I imagine adding multiple mints would be a normal possible use-case, although probably of different assets. Would the error only be if the same policy_id:asset_name was already present? Summing also seems like a reasonable behavior. I just don't like the old previous behavior of silently overwriting.

rooooooooob:
@gostkin could we merge this now? I made changes based on the comments you made.

gostkin (approved):

PR #217, opened 2023-05-26
WASM wrapper macros

rooooooooob:
creation for:
* to/from/ref traits for WASM wrappers to their rust base
* WASM lists (with elem being WASM-exposable) like what cddl-codegen creates
* WASM maps (with key/val being WASM-exposable) like what cddl-codegen creates


core/wasm/src/wasm_wrappers.rs
```diff
@@ -0,0 +1,131 @@
...
+        }
+    };
+}
+
+/// Convenience creator for generic WASM-exposable map
+/// This is exactly as the ones created by cddl-codegen
+/// so it is useful for utility functionality where those
+/// wouldn't have been automatically generated.
+#[macro_export]
+macro_rules! impl_wasm_map {
```
rooooooooob, line 84: all the types are separate since they might belong to different modules vs where this macro is called. I tried deriving the full path using a mod name + type and `paste` macro crate but it was too much hassle and there's not even any guarantee that the key/value types are in the same module (e.g. policyid in root, assetname in assets).

gostkin (approved):
lgtm

Issue #204, opened 2023-04-26
Fix building from the root folder


gostkin:
genesis module is missing

rooooooooob:
Fixed by the following:  #204 #221 #222 #223

PR #220, opened 2023-05-31
Misc compilation fixes, esp for WASM but also genesis

rooooooooob:
Previous work on the builders/etc updates rust but introduced things caused the WASM to no longer compile. This updates the WASM bindings as well.

Includes genesis module that wasn't pushed before.

### DIFF
--- a/chain/rust/src/genesis/network_info.rs
+++ b/chain/rust/src/genesis/network_info.rs
@@ -1,0 +1,47 @@
+use crate::{
+    byron::ProtocolMagic,
+    plutus::CostModels
+};
+use cml_core::Int;
+
+#[derive(Debug, Clone, Eq, Hash, Ord, PartialEq, PartialOrd)]
+pub struct NetworkInfo {
+    network_id: u8,
+    protocol_magic: ProtocolMagic,
+}
+impl NetworkInfo {
+    pub fn new(network_id: u8, protocol_magic: ProtocolMagic) -> Self {
+        Self {
+            network_id,
+            protocol_magic,
+        }
+    }
+    pub fn network_id(&self) -> u8 {
+        self.network_id
+    }
+    pub fn protocol_magic(&self) -> ProtocolMagic {
+        self.protocol_magic
+    }
+
+    pub fn testnet() -> NetworkInfo {
+        NetworkInfo {
+            network_id: 0b0000,
+            protocol_magic: ProtocolMagic::from(1097911063)
+        }
+    }
+    pub fn mainnet() -> NetworkInfo {
+        NetworkInfo {
+            network_id: 0b0001,
+            protocol_magic: ProtocolMagic::from(764824073)
+        }
+    }
+}
+
+// TODO: https://github.com/dcSpark/cardano-multiplatform-lib/issues/92
+pub fn plutus_alonzo_cost_models() -> CostModels {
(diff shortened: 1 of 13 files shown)


rooooooooob:
note: actually has a couple compile errors still that I noticed when trying from a clean repo. These are all addressed in the WASM wrapper work though and lets both PRs be up without conflicts.

rooooooooob:
This will fix #204 but only once #221 #222 #223 are merged (maybe not all are necessary but it definitely compiles with all together).

chain/rust/src/genesis/network_info.rs
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+    pub fn network_id(&self) -> u8 {
+        self.network_id
+    }
+    pub fn protocol_magic(&self) -> ProtocolMagic {
+        self.protocol_magic
+    }
+
+    pub fn testnet() -> NetworkInfo {
```
gostkin, line 26: what do you think about calling it preprod / preview? there're multiple testnets

chain/rust/src/genesis/network_info.rs
```diff
@@ -0,0 +1,47 @@
...
+        NetworkInfo {
+            network_id: 0b0000,
+            protocol_magic: ProtocolMagic::from(1097911063)
+        }
+    }
+    pub fn mainnet() -> NetworkInfo {
```
gostkin, line 32: [not important] from formatting perspective would be great to have an empty line before function signature

chain/rust/src/lib.rs
```diff
@@ -122,7 +122,7 @@ pub type DeltaCoin = Int;
 
 pub type GenesisHashList = Vec<GenesisHash>;
 
-pub type NetworkId = u8;
+pub type NetworkId = u32;// TODO: u8? or custom struct? (u8 can't be exposed to wasm)
```
gostkin, line 125: if we don't encode type size anywhere we could probably live with u32, what do you think?

gostkin (changes requested):
pls tell what you think about the review comments 🙏🏻

chain/rust/src/lib.rs
rooooooooob: It could give an API for testing against/for specific networks. e.g. `NetworkId::mainnet()`. Not all u32 values (in fact most wouldn't) work here either. That's easily addressed in the deserialization code but for setting it forces everything to return an error where you pass in a `u32`.

chain/rust/src/genesis/network_info.rs
rooooooooob: I just took this struct as-is from the old code, just with the wasm_bindgen part removed. @SebastienGllmt what should we do here? Aren't these for byron/genesis support? (hence the protocol magic, etc)

chain/rust/src/lib.rs
gostkin: I'd play it simple and have what can be easily exposed into wasm. in any case if we provide a couple of notes that network id is usually 0 or 1 people will figure out how to work with this

chain/rust/src/genesis/network_info.rs
gostkin: in any case protocol magics for preview and preprod are different. as we are creating an sdk I think we should support all testnets

chain/rust/src/genesis/network_info.rs
rooooooooob: This file was copied from the old code-base (just removing the wasm-bindgen parts). I'll add one in when we modify it for the testnet stuff.

chain/rust/src/genesis/network_info.rs
rooooooooob: Are they the same though? In this struct, used for genesis/byron, it's 32-bits and it can definitely use all of those. Elsewhere we have 4-bits of data in addresses, referred to as "network id" so I'm not sure if this is completely unrelated. I'm completely ignorant about protocol magic after byron though so I thought it was replaced with just the network id. The binary spec makes no reference to protocol magics, only network id.

What are the protocol magics for current networks?

rooooooooob: Ah, is this about CIP34? We could implement the stuff in that registry then I guess. It's not directly part of the on-chain data (post-byron) so we should document what it's referring to I guess. If we do that it might make more sense to move it away from the genesis module then.

rooooooooob:
@gostkin  what if we leave it how it is in this PR, since its main purpose is to get the crate compiling/working, and then after in a later PR move the network info + network ID into something like `network.rs` or `network_info.rs` in the root crate if we intend to support network info for current networks? (e.g. merging this byron network info with something that also contains info for current networks using CIP34's network registry) Or do we want to keep this (which is for legacy byron support) separate and clearly document that?

gostkin (approved):

PR #222, opened 2023-05-31
WASM code for Byron + updated CDDL to match types used

### LINKED ISSUE #213: Byron WASM
WASM modules are missing the byron work previously done to the rust code

rooooooooob:
Fixes #213

### DIFF
--- a/chain/wasm/src/byron/mod.rs
+++ b/chain/wasm/src/byron/mod.rs
@@ -1,0 +1,546 @@
+#![allow(
+    clippy::len_without_is_empty,
+    clippy::too_many_arguments,
+    clippy::new_without_default
+)]
+// This file was code-generated using an experimental CDDL to rust tool:
+// https://github.com/dcSpark/cddl-codegen
+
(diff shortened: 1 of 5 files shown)


chain/wasm/src/byron/mod.rs
```diff
@@ -0,0 +1,546 @@
+#![allow(
```
gostkin, line 1: should be added to lib.rs

rooooooooob: They already are. I can remove it here if that covers everywhere in the crate. cddl-codegen spits this out in the lib.rs of the output crate, it's just that we put that output crate into the byron mod here and rename it to mod.rs.

gostkin (approved):

chain/wasm/src/byron/mod.rs
gostkin: I meant not the `#![allow(` thing. there's no `mod byron` in `chain/wasm/src/lib.rs`

PR #223, opened 2023-05-31
WASM code for builders (all)

### LINKED ISSUE #203: WASM bindings for builders

rooooooooob:
Includes direct 1:1 wrappers for misc builders (#197) and tx_builder (#210)

Fixes #203

### DIFF
--- a/chain/wasm/src/builders/witness_builder.rs
+++ b/chain/wasm/src/builders/witness_builder.rs
@@ -1,0 +1,227 @@
+use wasm_bindgen::prelude::{wasm_bindgen, JsError};
+use cml_core_wasm::impl_wasm_conversions;
+use cml_crypto_wasm::{
+    DatumHash, Ed25519KeyHash, ScriptHash
+};
+use crate::{
+    Script,
+    byron::ByronAddress,
+    crypto::{BootstrapWitness, Vkeywitness},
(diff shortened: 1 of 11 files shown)


chain/wasm/src/builders/witness_builder.rs
```diff
@@ -0,0 +1,227 @@
...
+
+#[wasm_bindgen]
+#[derive(Debug, Clone)]
+pub struct PlutusScriptWitness(cml_chain::builders::witness_builder::PlutusScriptWitness);
+
+impl_wasm_conversions!(cml_chain::builders::witness_builder::PlutusScriptWitness, PlutusScriptWitness);
+
+#[wasm_bindgen]
+impl PlutusScriptWitness {
+    // pub fn script(&self) -> Option<PlutusScript> {
```
gostkin, line 25: why commented?

gostkin:
overall lgtm

chain/wasm/src/builders/witness_builder.rs
rooooooooob: It's how the rust part was. I think it slipped through before and we didn't end up needing to port it over in the utility code. I can either write the appropriate function in the rust builder (and add the wrapper here), or just remove the commented out function from both.

Issue #209, opened 2023-05-03
Support duplicate keys before Valentine


SebastienGllmt:
The Valentine hardfork removed the ability to have duplicate keys in map types in Cardano (https://github.com/dcSpark/cddl-codegen/issues/38). Theoretically CML should support duplicate keys before Valentine (at least in key places where they actually occurred onchain on mainnet) to be fully compatible

rooooooooob:
Do we know all the places we need to support? We need to in metadatum keys (or was it metadata labels?). Did we ever encounter it in plutus datums?

SebastienGllmt:
I think right now in CML we have patches for duplicate keys in all the places we ran into it in practice, right?

SebastienGllmt:
https://github.com/dcSpark/cardano-multiplatform-lib/pull/74

PR #226, opened 2023-06-02
Handle duplicate tx metadata labels

rooooooooob:
Addresses part of #209. Further testing needed to see which other cases exist on-chain.

### DIFF
--- a/core/rust/src/metadata.rs
+++ b/core/rust/src/metadata.rs
@@ -1,15 +1,147 @@
 use crate::error::{DeserializeError, DeserializeFailure, Key};
 use crate::ordered_hash_map::OrderedHashMap;
-use crate::serialization::{Deserialize, LenEncoding, Serialize, StringEncoding};
+use crate::serialization::{Deserialize, LenEncoding, Serialize, StringEncoding, fit_sz};
 use crate::Int;
 use cbor_event::{de::Deserializer, se::Serializer};
 use derivative::Derivative;
 
+use std::collections::BTreeMap;
 use std::io::{BufRead, Seek, SeekFrom, Write};
 
 pub type TransactionMetadatumLabel = u64;
 
-pub type Metadata = OrderedHashMap<TransactionMetadatumLabel, TransactionMetadatum>;
+/// Collection of TransactionMetadatums indexed by TransactionMetadatumLabels
+/// Handles the extremely rare edge-case of in previous generations allowing
+/// duplicate metadatum labels.
+#[derive(Clone, Debug, Default, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+pub struct Metadata {
+    pub entries: Vec<(TransactionMetadatumLabel, TransactionMetadatum)>,
+    #[serde(skip)]
+    pub encodings: Option<MetadataEncoding>,
+}
+
+impl Metadata {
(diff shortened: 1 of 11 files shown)


core/rust/src/metadata.rs
```diff
@@ -1,15 +1,147 @@
...
 
 pub type TransactionMetadatumLabel = u64;
 
-pub type Metadata = OrderedHashMap<TransactionMetadatumLabel, TransactionMetadatum>;
+/// Collection of TransactionMetadatums indexed by TransactionMetadatumLabels
+/// Handles the extremely rare edge-case of in previous generations allowing
+/// duplicate metadatum labels.
+#[derive(Clone, Debug, Default, serde::Deserialize, serde::Serialize, schemars::JsonSchema)]
+pub struct Metadata {
+    pub entries: Vec<(TransactionMetadatumLabel, TransactionMetadatum)>,
```
rooooooooob, line 18: We don't ever expect to see massively combined metadatas with like 1000+ labels, right?

It should be faster this way for reasonable amounts of labels, and you need to remember the order 100% so something like `OrderedHashMap<TransactionMetadatumLabel, Vec<TransactionMetadatum>>` wouldn't work as you could have something like `100`, `50`, `100` which would turn it into `100`, `100`, `50`.

rooooooooob: Of course you could do a hybrid approach and only remember it that way in the `MetadataEncodings` if we do need to support massive amounts of labels.

SebastienGllmt: As in 1000s of collisions? No.

Also, this PR is related to https://github.com/dcSpark/cardano-multiplatform-lib/issues/209

rooooooooob: Yeah #209 is in the description but I didn't put "fixes" or anything since it might not cover all the cases of duplicate keys in cbor in general. I meant 1000s of metadatum labels in general in a single auxdata. Presumably different ones. i.e. if a hybrid storage option is needed (with Vec in the `MetadataEncodings` to preserve order 100% even for duplicates mixed in but a Map in `Metadata` itself) for faster lookup by label.

SebastienGllmt (approved):
LGTM other than the fact that we can re-use the old code for the post-Valentine data format since that one can't contain duplicate labels

gostkin (approved):

gostkin:
> LGTM other than the fact that we can re-use the old code for the post-Valentine data format since that one can't contain duplicate labels

@SebastienGllmt I think what @rooooooooob did will work well with post-Valentine data, since technically you can't insert 2 same keys (using the methods). the only thing we might probably need to change is public access to fields in `Metadata` struct. this way it would be impossible to mess up the structure with the provided api, while it would still work with old data within the same struct

rooooooooob:
IIRC we wanted to have metadata as a separate struct at some point anyway to implement traits on it more easily. We could move this to the legacy multiera parsing types when we move over to conway. We could either make another type like this just without duplicate support (e.g. just wrapping `OrderedHashMap<Label, Datum>`) or we could use the `OrdereedHashMap` itself directly like we did before. We would have had to regenerate anything affected by changing Metadata/auxdata regardless since tx body itself changes in `conway.cddl` vs `babbage.cddl`, thus tx/bocks/etc too.

PR #227, opened 2023-06-06
Update/fix tests (Address/Byron)

### LINKED ISSUE #219: Fix address parsing issue
Byron test failing. Base58 representation incorrect. Looks like inner CBOR is different.

rooooooooob:
Update byron tests

Update address tests + fix for byron address serialization (old code would ommit bootstrap in addr attributes contrary to the stated byron.cddl)

Fix doc-tests in crypto lib

Fixes #219

### DIFF
--- a/chain/rust/src/byron/serialization.rs
+++ b/chain/rust/src/byron/serialization.rs
@@ -18,10 +18,13 @@ impl cbor_event::se::Serialize for AddrAttributes {
         &self,
         serializer: &'se mut Serializer<W>,
     ) -> cbor_event::Result<&'se mut Serializer<W>> {
+        // NOTE: This was manually modified *slightly*
+        // The direct cddl-codegen code will always include the stake distribution
+        // but the old byron code seems to only put it in when it's not bootstrap.
         serializer.write_map(cbor_event::Len::Len(
             match &self.stake_distribution {
-                Some(_) => 1,
-                None => 0,
+                Some(StakeDistribution::SingleKey(_)) => 1,
+                _ => 0,
             } + match &self.derivation_path {
                 Some(_) => 1,
                 None => 0,
@@ -30,10 +33,10 @@ impl cbor_event::se::Serialize for AddrAttributes {
                 None => 0,
             },
         ))?;
-        if let Some(field) = &self.stake_distribution {
+        if let Some(StakeDistribution::SingleKey(_)) = &self.stake_distribution {
             serializer.write_unsigned_integer(0u64)?;
             let mut stake_distribution_inner_se = Serializer::new_vec();
-            field.serialize(&mut stake_distribution_inner_se)?;
+            self.stake_distribution.as_ref().unwrap().serialize(&mut stake_distribution_inner_se)?;
             let stake_distribution_bytes = stake_distribution_inner_se.finalize();
             serializer.write_bytes(&stake_distribution_bytes)?;
         }
--- a/chain/rust/src/address.rs
+++ b/chain/rust/src/address.rs
@@ -858,12 +858,13 @@ impl Deserialize for RewardAccount {
         }
     }
 }
-/*
+
 #[cfg(test)]
 mod tests {
-    use crate::{byron::AddressContent, ledger::common::hash::ScriptHashNamespace};
+    use crate::{byron::AddressContent, transaction::NativeScript};
 
     use super::*;
+    use cml_core::serialization::ToBytes;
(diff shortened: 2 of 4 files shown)


chain/rust/src/byron/serialization.rs
```diff
@@ -18,10 +18,13 @@ impl cbor_event::se::Serialize for AddrAttributes {
         &self,
         serializer: &'se mut Serializer<W>,
     ) -> cbor_event::Result<&'se mut Serializer<W>> {
+        // NOTE: This was manually modified *slightly*
```
rooooooooob, line 21: @SebastienGllmt just to confirm is this the desired behavior? The tests fail without it and it's what the code in the old byron/serialization.rs has, but is not what the direct byron.cddl does. I completely skimmed over it in the old code and thought I was going crazy when debugging and comparing vs old code.

SebastienGllmt: Could be a mistake. I tried to manually check to make sure everything matched since the old byron.cddl was a mess and I had to fix multiple things in it, but maybe I missed this

rooooooooob: I assume it was on purpose. The code would have been manually edited after codegen before to be like that. It doesn't start in `Ae2` like byron addresses used to IIRC, but weren't there multiple kinds? (Icarus or not). I don't know as much about Byron as you do though so I'm not sure. It's possible that the change should be, instead of being in serialization, in the construction to avoid populating the stake distribution if it's bootstrap instead of including it in the first place. The one created by including the bootstrap stake distribution is `2657WMsDfac5izkoz9CMcGTDcLwq5KGcNcj5GT2HMxvy11CCNjj2ae7PvEAcqVivB`. Does that look like any kind of byron address? (non-icarus or whatever)

SebastienGllmt: FYI this is what is done in the old Rust implementation of Byron by IOG https://github.com/input-output-hk/rust-byron-cardano/blob/5f0c8d34424e0ccb199b8c6b70fd410d4bd6af41/cardano/src/address.rs#L257

The spec also has some mistakes in it, and you can see me talking about them here https://github.com/input-output-hk/cardano-ledger/pull/2923

SebastienGllmt: `2657WMsDfac5izkoz9CMcGTDcLwq5KGcNcj5GT2HMxvy11CCNjj2ae7PvEAcqVivB` Daedalus addresses usually start with Ae2 or Dd, but there are two cases where this isn't the case

1. On testnet, they have different prefixes (I think one of them started with 2?). Note Cardano didn't have testnet support at first, then it was added later (it was made backwards compatible by defining mainnet as having no protocol magic in the address)
2. Some Byron addresses have junk data stuffed into them (since note the address contains a map you can just put anything you want into them theoretically). This was then disallowed in a later version of Cardano

rooooooooob: It looks like that Rust implementation of Byron does the same thing (omits it entirely if bootstrap). This wouldn't be 2) since the only extra information was the bootstrap stake distribution. I don't think it's 1) either since the address is for mainnet.

gostkin: @rooooooooob let's add some testcases for these addresses starting with 2 and also for the long addresses

gostkin: could be probably in old sdk issues

rooooooooob: @gostkin @SebastienGllmt Do we need test cases for these before we can merge this PR? I tried trawling through old issues/PRs and I found workarounds in input-output-hk/cardano-byron-cli#73 and Emurgo/project-icarus-importer#3 but no addresses are explicitly mentioned (also searching the block mentioned there seems to give normal address lengths - maybe the explorer trims them?)

I could sync through byron again and explicitly check but I don't think it's worth it. The addresses are _**huge**_ and there's no difference encoding-wise. There's just a plain `bytes` field inside the byron address' HD payload that has zero restrictions and can contain anything. I don't see what the test would accomplish besides maybe lagging your editor when you open `address.rs`. We've successfully round-tripped every byron block so I assume there's no issues.

I could add in a testnet one though, that should be easier to find.

rooooooooob: Actually it looks like we already have tests in `chain/rust/src/byron/utils.rs` to cover testnet addresses.

PR #232, opened 2023-06-27
Unused imports removed

rooooooooob:
* Remove unused imports

* Some tx_builder.rs tests weren't unwrapping add_mint() calls to check for Err case

* Some other unused function/variable (outside codegened code) warnings fixed too

### DIFF
--- a/chain/rust/src/byron/utils.rs
+++ b/chain/rust/src/byron/utils.rs
@@ -153,9 +153,9 @@ impl AddressContent {
     }
 
     // icarus-style address (Ae2)
-    pub fn icarus_from_key(key: Bip32PublicKey, protocol_magic: u32) -> AddressContent {
+    pub fn icarus_from_key(key: Bip32PublicKey, protocol_magic: ProtocolMagic) -> AddressContent {
         // need to ensure we use None for mainnet since Byron-era addresses omitted the network id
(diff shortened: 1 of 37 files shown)


chain/rust/src/byron/utils.rs
```diff
@@ -153,9 +153,9 @@ impl AddressContent {
     }
 
     // icarus-style address (Ae2)
```
rooooooooob, line 155: Note: I started this PR thinking that those 6 simultaneous PRs had a compile error after merging them together and this would be to fix that, but after realized this fix is in #227 and it just wasn't merged yet. It's the same change here/there so it should be fine though. Just had to change the param type to fix the WASM bindings. The rest of the PR is just fixing warnings though.

rooooooooob:
Besides that one protocol magic thing this is just removing imports and a couple fixes for warnings so we should be able to merge this quickly.

rooooooooob:
@gostkin  Besides that one protocol magic thing this is just removing imports and a couple fixes for warnings so we should be able to merge this quickly.

rooooooooob:
@gostkin I'm just going to merge this so I can put up more PRs for multi-era without any merge conflicts. Let me know if there's anything changed here that needs changes. The behavior didn't really change so it's probably fine.

PR #246, opened 2023-07-20
Change Babbage block header encoding

rooooooooob:
While all other eras have the header use basic encoding for the operational cert + protocol version (i.e. inline the fields inside), the babbage blocks we've seen seem to not follow the CDDL spec and instead treat them as if they were not basic groups but instead array groups.


rooooooooob:
I made an issue for the CDDL difference here input-output-hk/cardano-ledger#3559

gostkin (approved):
lgtm, thanks for reporting the issue

PR #247, opened 2023-07-25
WASM build fixes

rooooooooob:
* Some WASM list/map wrappers were redundant in cml-chain/multi-era resulting in duplicate JS types

* Mix of `to_js_value()` and `to_json_value()` depending when we generated the types, which breaks the JSON schema script.

* Regen WASM using dcSpark/cddl-codegen#200 to massively reduce boilerplate and unify the JSON/CBOR conversiosn so we only need to change one place (this fixes the above point)

* wasm/json-gen crates were broken (incorrect module resolution or referring to types we removed)

* JSON schema build script replaces correctly for enum cases now, and also
fixes the type names fore recursive types that had 1's appended to them.

* Added in exports for missing types in chain/cip25 json export crates

### DIFF
--- a/scripts/run-json2ts.js
+++ b/scripts/run-json2ts.js
@@ -102,7 +102,9 @@ Promise.all(schemaFiles.map(schemaFile => {
   // prepend 'JSON' to all identifiers here so they don't conflict with main .ts types
   for (let i = 0; i < dedupedDefs.length; ++i) {
     for (let id of added) {
-      dedupedDefs[i] = dedupedDefs[i].replace(new RegExp(`\\b${id}\\b`), id + 'JSON');
+      // 1) To avoid FooJSON: FooJSON instead of a more consistent Foo: FooJSON we don't replace if followed by :
+      // 2) For some reason we're getting recursive types having a 1 appended to the type so cover that too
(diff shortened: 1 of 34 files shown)


scripts/run-json2ts.js
```diff
@@ -102,7 +102,9 @@ Promise.all(schemaFiles.map(schemaFile => {
   // prepend 'JSON' to all identifiers here so they don't conflict with main .ts types
   for (let i = 0; i < dedupedDefs.length; ++i) {
     for (let id of added) {
-      dedupedDefs[i] = dedupedDefs[i].replace(new RegExp(`\\b${id}\\b`), id + 'JSON');
+      // 1) To avoid FooJSON: FooJSON instead of a more consistent Foo: FooJSON we don't replace if followed by :
+      // 2) For some reason we're getting recursive types having a 1 appended to the type so cover that too
```
SebastienGllmt, line 106: Did we ever figure out the "For some reason" here?

SebastienGllmt (approved):
Great idea

scripts/run-json2ts.js
rooooooooob: No but I didn't look into their implementation or anything. We had problems even getting recursive types to work in the first place so maybe this is some workaround the json schema-related tools we call do to get around it, but it's been a long time so I don't remember the specifics. It's an issue for tx metadata + plutus datums as those are our only recursive types.

PR #250, opened 2023-08-09
Multi era json and lint fixes + CI

### LINKED ISSUE #234: impl_wasm_map! unaware of primitive vs not
impl_wasm_map! needs to be aware of if the keys/values are primitives or not as wasm_bindgen does not allow references for primitives, and CML uses references for non-primitives for consistency + to avoid confusing crashes (it moves the params in so if you refer to it later from JS it crashes which is confusing since unlike rust there's no compile to check that you're not using a moved value and 99% of people trying to use the library are probably completely unaware of this).

### LINKED ISSUE #245: json-gen for dependent crates
e.g. multi-era's JSON depends on chain's JSON so we need to test this as it won't find the schemas from chain as those aren't exported as part of the build

rooooooooob:
* Add package.json for multi-era wasm building

* Fix up shelley move inst rewards + certs (they are slightly more
restrictive than post-shelley ones that were being used before)

* Change chain/multi-era json-gen crates to be usable as a library too so
we can export all cml-chain JSON schemas from cml-multi-era

* Warnings / clippy lint fixes

* Adds CI for building/testing new crate structure

* impl_wasm_map is now aware of whether the keys/values are WASM-exposable
and copy or not. There is now a away to specify the backing map e.g.
BTreeMap for preserve-encodings=false i.e. Byron, CIP25

* Use of this macro now replaces hard-generated WASM Map (and list) types
to reduce boilerplate + allow a single common place to fix linting
errors (one of the main reasons for this change to be done in this PR -
but it was planned at some point anyway). The rest of clippy linting errors were also fixed and a check in the CI
for fmt + linting.

Fixes #234

Fixes #245


rooooooooob:
~TODO: There are a few clippy warnings to fix. Will fix those and add a cargo clippy lint / fmt CI step when I do.~

**Edit: Done**

.github/workflows/pr-checks.yml
```diff
@@ -42,3 +42,41 @@ jobs:
       - name: rust:build-browser
         run: |
           npm run rust:build-browser
+      - name: all:test
```
gostkin, line 55: probably it makes sense to have multiple checks, e.g. why to test old cml when only new crates are changed? you can specify the path for particular checks

gostkin (approved):

PR #251, opened 2023-08-30
Update CIP36 to changes to spec

rooooooooob:
Also updated test vectors

WASM utils were added as they were left as a TODO from before

### DIFF
--- a/cip36/rust/src/lib.rs
+++ b/cip36/rust/src/lib.rs
@@ -190,9 +190,10 @@ mod tests {
         // test case says: 0036ef3e1f0d3f5989e2d155ea54bdb2a72c4c456ccb959af4c94868f473f5a0
         // but the CBOR differs, so we use the ones in the CBOR for the test case.
         // Is this being derived or why does it not match?
-        let stake_cred = PublicKey::from_raw_bytes(
-            &[227, 205, 36, 4, 200, 77, 230, 95, 150, 145, 143, 24, 213, 180, 69, 188, 185, 51, 167, 205, 161, 142, 237, 237, 121, 69, 221, 25, 30, 67, 35, 105]
-        )
+        let stake_cred = PublicKey::from_raw_bytes(&[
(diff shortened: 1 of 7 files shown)


cip36/rust/src/lib.rs
```diff
@@ -187,63 +187,66 @@ mod tests {
         //     .derive(0)
         //     .derive(0);
         // let catalyst_prv_key = hex::decode("4820f7ce221e177c8eae2b2ee5c1f1581a0d88ca5c14329d8f2389e77a465655c27662621bfb99cb9445bf8114cc2a630afd2dd53bc88c08c5f2aed8e9c7cb89").unwrap();
-        let stake_cred = PublicKey::from_raw_hex(
-            "86870efc99c453a873a16492ce87738ec79a0ebd064379a62e2c9cf4e119219e",
+        // test case says: 0036ef3e1f0d3f5989e2d155ea54bdb2a72c4c456ccb959af4c94868f473f5a0
```
rooooooooob, line 190: test vectors: https://github.com/cardano-foundation/CIPs/blob/master/CIP-0036/test-vector.md

I had issues with the stuff in the human readable not matching what's in the serialized CBOR the first time I made the CIP36 with the old test vectors too. Not sure if there's some deriving being done or if they just don't match.

rooooooooob: You can see:
```
Derivation path

To avoid linking voting keys directly with Cardano spending keys, the voting key derivation path must start with a specific segment:

m / 1694' / 1815' / account' / chain / address_index

We recommend that implementors only use address_index=0 to avoid the need for voting key discovery.
```

but IIRC this didn't make them match either. I didn't check it out here but I could look into it once conway/etc is done. The CBOR portion of the lib is fine.

gostkin (approved):

gostkin:
@rooooooooob pls fix the linters

PR #256, opened 2023-09-01
added docs folder with docusaurus project

nhaga:


docs/docs/enums.mdx_bak
```diff
@@ -0,0 +1,135 @@
+# Enums
+
+## Certificate
+
```
rooooooooob, line 4: I'm confused as to the point of this file. It seems more like something that should be covered by something automatically generated like what crates.io does with the rust API.

docs/docs/index.md
```diff
@@ -0,0 +1,57 @@
...
+
+You can find the specifications of Cardano's ledger [here](https://github.com/input-output-hk/cardano-ledger-specs) which we suggest consulting as you use this library. Notably, the `Shelley ledger formal specification` covers the core concepts.
+
+
+## Benefits of using this library
+
+Serialization/deserialization code is automatically generated from
+Cardano’s official specification, which guarantees it can easily stay up
+to date! We do this using a tool managed by EMURGO & dcSpark called `cddl-codegen`_
+which can be re-used for other tasks such as automatically generate a
```
rooooooooob, line 40: >such as automatically generate a

Grammar. I know it was probably wrong wherever it was originally from in the old docs but we might as well fix it here.

docs/docs/index.md
```diff
@@ -0,0 +1,57 @@
...
+It is also very easy to create scripts in Rust or WASM to share with
+stake pools, or even embed inside an online tool! No more crazy
+cardano-cli bash scripts!
+
+Powerful and flexible enough to be used to power wallets and exchanges!
+(Yes, it’s used in production!)
+
+## Documentation
+
+This library generates both `Typescript`_ and `Flow`_ type definitions,
```
rooooooooob, line 52: We don't use flow anymore in the new CML

docs/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "cddl-codegen-documentation",
```
rooooooooob, line 2: cddl-codegen? (I assume a copy-paste error?)

docs/docusaurus.config.js
```diff
@@ -0,0 +1,171 @@
...
+        //         to: "/blog",
+        //       },
+        //       {
+        //         label: "GitHub",
+        //         href: "https://github.com/facebook/docusaurus",
+        //       },
+        //     ],
+        //   },
+        // ],
+        copyright: `Copyright © ${new Date().getFullYear()} Milkomeda Foundation. Built with Docusaurus.`,
```
rooooooooob, line 160: Milkomeda Foundation?

docs/docs/type_definitions.mdx_bak
```diff
@@ -0,0 +1,25 @@
+# Type Definitions
+
+## PolicyID
+
+```rust
+pub type PolicyID = ScriptHash;
+```
+
```
rooooooooob, line 8: see comment in enums section

docs/docs/modules/metadata.mdx
```diff
@@ -0,0 +1,400 @@
...
+  // structure did not match
+}
+```
+
+For decoding in a more exploratory manner we can check the types first as such:
+```javascript
+function parseMetadata(metadata) {
+  // we must check the type first to know how to handle it
+  switch (metadata.kind()) {
+    case CardanoWasm.TransactionMetadatumKind.MetadataMap:
```
rooooooooob, line 122: this code will need to be updated. the variants here are called `Map` and `List` now, not `MetadataMap`.

docs/docs/modules/builders/generating_transactions.mdx
```diff
@@ -0,0 +1,121 @@
...
+const transaction = CardanoWasm.Transaction.new(
+    txBody,
+    witnesses,
+    undefined, // transaction metadata
+);
+```
+
+## A note on fees
+
+Fees is Cardano Shelley are based directly on the size of the final encoded transaction. It is important to note that a transaction created by this library potentially can vary in size compared to one built with other tools. This is because transactions, as well as other Cardano Shelley structures, are encoded using [CBOR](https://cbor.io/) a binary JSON-like encoding. Due to arrays and maps allowing both definite or indefinite length encoding in the encoded transaction created by the library, the size can vary. This is because definite encoding consists of a tag containing the size of the array/map which can be 1 or more bytes long depending on the number of elements the size of the encoded structure, while indefinite length encoding consists of a 1 byte starting tag and after all elements are listed, a 1 byte ending tag. These variances should should only be a couple bytes and cardano-multiplatform-lib uses definite encoding which is the same length or smaller for any reasonable sized transaction.
```
rooooooooob, line 121: We do match encodings now (when creating via CBOR bytes) but will use definite encodings by default still so maybe this passage is still relevant. It could be worth mentioning this at least though. The "Cardano Shelley" wording as well. I know this was just pulled from the old docs to get the format in but we'll want to look through these moved-over things and make updates like this.

docs/docs/modules/builders/generating_transactions.mdx
```diff
@@ -0,0 +1,121 @@
...
+    CardanoWasm.BigNum.from_str('44'),
+    CardanoWasm.BigNum.from_str('155381')
+);
+const txBuilderCfg = CardanoWasm.TransactionBuilderConfigBuilder.new()
+    .fee_algo(linearFee)
+    .pool_deposit(CardanoWasm.BigNum.from_str('500000000'))
+    .key_deposit(CardanoWasm.BigNum.from_str('2000000'))
+    .max_value_size(4000)
+    .max_tx_size(8000)
+    .coins_per_utxo_word(CardanoWasm.BigNum.from_str('34482'))
```
rooooooooob, line 40: the API has been changed. this code must've been from a previous era as we don't have coins_per_utxo_word anymore. There's a coins_per_utxo_byte but I'm guessing there are other areas of this code that needs updating too.

rooooooooob (changes requested):
A few minor changes need to be done (see comments - copy-paste errors + a grammar issue).

I don't think there's much utility in the enums / types files so they could be removed.

I know it's probably just moving over old docs to the new format to get a starting structure, but we'll need (in a later PR?) to update the content of the docs that were copy pasted over from the old docs. Most of those were written in the Shelley / Allegra / Mary eras and some things have changed for Alonzo onward.

rooooooooob (approved):
I'm just going to approve and merge this PR and do the fixes in a new branch since this is from a fork.

The failing check is happening on `develop` too and is fixed in #258

PR #257, opened 2023-09-02
Derive Copy for types based on fixed-len arrays

oskin1:


rooooooooob (changes requested):
Sorry for the late review. Thanks for the PR. Could you fix the linting errors? Since it's `Copy` now you'll end up with some places with unnecessary clones now which will trigger the clippy CI job. Try rebasing off `develop` if you only want to fix the ones caused by this change (there are a couple warning ones that were fixed on `develop` here but might cause an issue on this branch if you don't)

PR #258, opened 2023-09-05
Conway initial cml-chain update

rooooooooob:
Includes cml-chain/cml-chain-wasm's structs except for the builders.

Util functionality (besides builders) were updated.

The builders are temporarily disabled and will be updated in a coming commit.

Also needs JSON stuff moved over and clippy warnings to be fixed again.

cml-multi-era will need updating too to bring over babbage and to update existing structs (regen + update the cddls)


gostkin:
@rooooooooob pls fix the checks

chain/rust/src/builders/certificate_builder.rs
```diff
@@ -51,17 +52,23 @@ pub fn cert_required_wits(cert: &Certificate, required_witnesses: &mut RequiredW
...
             required_witnesses.add_vkey_key_hash(cert.ed25519_key_hash.clone());
         }
-        Certificate::GenesisKeyDelegation(cert) => {
-            required_witnesses.add_vkey_key_hash(
-                Ed25519KeyHash::from_raw_bytes(cert.genesis_delegate_hash.to_raw_bytes()).unwrap(),
-            );
-        }
-        // no witness as there is no single core node or genesis key that posts the certificate
-        Certificate::MoveInstantaneousRewardsCert(_cert) => {}
+        Certificate::RegCert(_cert) => todo!(),
```
gostkin, line 55: will we introduce that in the following prs?

chain/rust/src/transaction/utils.rs
```diff
@@ -17,48 +17,49 @@ impl TransactionOutput {
...
         }
     }
 
     pub fn amount(&self) -> &Value {
         match self {
-            Self::ShelleyTxOut(tx_out) => &tx_out.amount,
-            Self::AlonzoTxOut(tx_out) => &tx_out.amount,
-            Self::BabbageTxOut(tx_out) => &tx_out.amount,
+            Self::AlonzoFormatTxOut(tx_out) => &tx_out.amount,
+            Self::ConwayFormatTxOut(tx_out) => &tx_out.amount,
```
gostkin, line 45: babbage / shelley are subset of conway right?

chain/wasm/src/lib.rs
```diff
@@ -14,31 +14,39 @@ pub mod address;
 pub mod assets;
 pub mod auxdata;
 pub mod block;
-pub mod builders;
+//pub mod builders;
```
gostkin: 🙃

crypto/rust/src/lib.rs
```diff
@@ -626,6 +626,8 @@ impl_hash_type!(KESVkey, 32);
 // same for this signature (but lots of traits aren't implemented for [u8; 448] so we can't)
 //impl_hash_type!(KESSignature, 448);
 impl_hash_type!(NonceHash, 32);
+// not sure if this is an existing hash or not. made it a new one but maybe it's datum hash or something?
```
gostkin: 🤔

chain/rust/src/builders/certificate_builder.rs
rooooooooob: Yes

chain/rust/src/transaction/utils.rs
rooooooooob: CDDL:
```cddl
alonzo_format_tx_out = [
  address,
  amount : value,
  ? datum_hash
]
conway_format_tx_out = {
    0 : address,      ; @name address
    1 : value,        ; @name amount
  ? 2 : datum_option, ; @name datum_option
  ? 3 : script_ref,   ; @name script_reference
}

transaction_output = alonzo_format_tx_out / conway_format_tx_out
```

Starting from Alonzo they added the optional datum hash to the format (when left out it's the same as Shelley), and in Babbage they changed it completely changed it to be a map but kept the previous array format as well as it takes up less space on-chain (no need to store keys) + maybe some compatability concerns.

The reason I changed this for Conway is that the Babbage one contained a Babbage-era script ref (plutus v1/v2) whereas this one is a Conway-era one (plutus v1/v2/v3).

crypto/rust/src/lib.rs
rooooooooob: As far as I can tell it's its own new thing. The formal spec refers to it as a DocHash (hash of the contents at the url in the anchor pair). I'll rename it based on what I see in the formal spec for now, as that makes more sense than the name in the CDDL spec.

chain/wasm/src/lib.rs
rooooooooob: Fixed

gostkin (approved):

PR #264, opened 2023-10-02
Clarify where data is used in a redeemer

SebastienGllmt:
In some places, we use the variable name `plutus_data`, but unless you have the context it's not clear that this is actually the data for the redeemer and not the datum

I changed the variable names to try and clarify this


rooooooooob (commented):
What about the WASM bindings?

SebastienGllmt:
@rooooooooob can you maybe take over this PR?

PR #265, opened 2023-10-02
Not add inline datum data to tx

gostkin:
According to the ledger rules inline datums shouldn't be added to the transaction

### DIFF
--- a/chain/rust/src/builders/input_builder.rs
+++ b/chain/rust/src/builders/input_builder.rs
@@ -30,9 +30,6 @@ pub fn input_required_wits(
             }
             StakeCredential::Script { hash, .. } => {
                 required_witnesses.add_script_hash(hash.clone());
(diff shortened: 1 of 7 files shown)


SebastienGllmt (approved):
Makes sense to me assuming you tested that it works

rooooooooob (approved):
LGTM assuming I understood it, seems the ledger specs/CIP32 mention it:
>When an output with an inline datum is spent, the spending transaction does not need to provide the datum itself.

Do we want to do any sanity checking somewhere against mixing inline datums with Plutus V1 scripts? That was mentioned in a few places as not being allowed.

gostkin:
> Makes sense to me assuming you tested that it works

checked with this PR https://github.com/dcSpark/projected-nft-whirlpool/pull/11

PR #267, opened 2023-10-09
Doc updates + example dApp + misc WASM functionality

rooooooooob:
Added content to docs / misc fixes

Added an example dApp that uses cml-chain + cml-cip25 to mint a NFT via connecting to Flint

Various functionality missing in WASM bindings was added that was noticed while implementing this dApp


gostkin (commented):
I can't find the dapp :)

rooooooooob:
Oops! I forgot to add them to the commit. I just pushed a commit with the directory. @gostkin

gostkin (changes requested):
the CI needs to be fixed, also would be good to add a CI for the example

rooooooooob:
@gostkin  I have no idea why the CI is failing. I tried looking into it yesterday. It's happening in another open PR too. It's not related to this PR at all yet the previously merged PRs had their CI pass. The failures are not producible locally for me or the other PR. The clippy error doesn't make any sense either as both `Ord` and `PartialOrd` are derived automatically where it's complaining about. I thought it could maybe be some update to the `derivative` crate with a minor version that is being used in the CI here but is a different minor version than my calls locally, but the crate hasn't been touched in a while.

gostkin (approved):

PR #268, opened 2023-10-13
Derive Copy for types based on fixed-len arrays

oskin1:
Duplicate of #257

### DIFF
--- a/chain/rust/src/address.rs
+++ b/chain/rust/src/address.rs
@@ -1363,7 +1363,7 @@ mod tests {
         let script_hash =
             ScriptHash::from_raw_bytes(oneof_native_script.hash().to_raw_bytes()).unwrap();
 
-        let spend_cred = StakeCredential::new_script(script_hash.clone());
+        let spend_cred = StakeCredential::new_script(script_hash);
         let stake_cred = StakeCredential::new_script(script_hash);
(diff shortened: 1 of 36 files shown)


oskin1:
@rooooooooob which version of clippy is used in by CI?

rooooooooob:
I'm not sure why the CI job run is failing. I can't reproduce it locally with the same command, and the clippy error doesn't make much sense as both `PartialOrd`/`Ord` are derived, albeit using the `derivative` crate. There doesn't seem to be any new version of the `derivative` crate possibly causing different behavior between the CI task and locally.

Either way it's definitely not related to this PR as it's happening in my other open PR but the last few merged PRs had their CI jobs complete just fine.

rooooooooob:
I put up a PR #269 that should fix it. I'll try and get it merged then rerun the check here and if it passes (it failing prevented a bunch of other later checks from running in this PR) I'll merge this.

rooooooooob:
@oskin1 I think you'll need to merge the latest `develop` into your branch to have the CI get passed that issue. I tried just making the CI run again but it still fails. I would just LGTM it and merge it but it would be good to make sure that all the other tests in the CI pass.

oskin1:
@rooooooooob rebased on top of develop

rooooooooob (approved):

PR #270, opened 2023-10-23
Conway cert builder support

rooooooooob:
support for adding conway certs to txs with support for deposit calculation

also fixes NetworkId bug

### DIFF
--- a/chain/rust/src/builders/certificate_builder.rs
+++ b/chain/rust/src/builders/certificate_builder.rs
@@ -25,24 +25,15 @@ pub enum CertBuilderError {
 // comes from witsVKeyNeeded in the Ledger spec
 pub fn cert_required_wits(cert: &Certificate, required_witnesses: &mut RequiredWitnessSet) {
     match cert {
-        // stake key registrations do not require a witness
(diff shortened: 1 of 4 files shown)


chain/rust/src/builders/certificate_builder.rs
```diff
@@ -101,7 +116,9 @@ pub fn add_cert_vkeys(
         Certificate::PoolRetirement(cert) => {
             vkeys.insert(cert.ed25519_key_hash.clone());
         }
-        Certificate::RegCert(_cert) => todo!(),
+        Certificate::RegCert(_cert) => {
+            // does not require a witness
+        }
         Certificate::UnregCert(_cert) => todo!(),
```
gostkin: why todos?

rooooooooob: Forgot to add them here too. The ones in cert_required_wits were all tested so this should be fine since it's based on those and identical except it errors when the wits aren't vkeys. I just pushed a commit that includes these too.

gostkin (approved):

PR #272, opened 2023-10-27
Multi-Era expanded functionality + Shelley Metadata set fix

rooooooooob:
All multi-era functionality besides some Byron-specific stuff should be exposed. e.g now the blocks can have the tx bodies/metadata/witnesses inspected and tx bodies can have all their fields inspected through era-agnostic wrappers. WASM wrappers are also provided.

ShelleyBlock's metadataset was not wrapped in the tx index map for some reason. This still passed the deserialization tests since this still ended up being able to parse the metadata as the metadatum label map/tx index map are the same and then the inner label was also valid regular metadata by itself. After the fix mainnet shelley was retested just in case and there were no issues.


multi-era/rust/src/utils.rs
```diff
@@ -60,6 +98,1204 @@ impl MultiEraBlock {
...
+    pub fn block_number(&self) -> Option<u64> {
+        match self {
+            Self::ByronEB(_) => None,
+            Self::Byron(_) => None,
+            Self::Shelley(header) => Some(header.body.block_number),
+            Self::Babbage(header) => Some(header.header_body.block_number),
+        }
+    }
+
+    pub fn slot(&self) -> Option<u64> {
```
rooooooooob, line 256: Do we want to do something for Byron? I'm not sure how the transition went there. IIRC it used to be like 22600 or something slots per epoch, and the slots reset but I don't really know/remember. Should we calculate it like that? Or is that only on mainnet that it was that number? Or should we just only have this work on shelley-onward blocks?

gostkin: need to check

gostkin:
just in case is there a way to iterate through e.g. all tx inputs or outputs without parsing each block type?

gostkin:
would be good to have numeric era rep as well, wdyt?

rooooooooob:
>would be good to have numeric era rep as well, wdyt?

Do you mean like the one corresponding to what pallas gives like 6 for babbage, 5 for alonzo, etc?

>just in case is there a way to iterate through e.g. all tx inputs or outputs without parsing each block type?

I'm not sure if I understand. Like parsing only part of the CBOR bytes? You'd need to parse everything before it to know when they start as otherwise you wouldn't know how many bytes to skip. You'd be able to skip parsing the witnesses in the blocks since the tx bodies come before that. It would be pretty mess though.

gostkin:
> > would be good to have numeric era rep as well, wdyt?
> 
> Do you mean like the one corresponding to what pallas gives like 6 for babbage, 5 for alonzo, etc?
> 
yes

rooooooooob:
You would only be able to get that info if you parsed it from that specific format. It's impossible to know otherwise since it's not stored anywhere in the actual block CBOR. Are you recommending storing it as an `Option` and only returning it for when it was parsed from the pallas bytes and returning `None` otherwise? Otherwise I could make it check the format and return it based just on that but then you'd have it reporting `Shelley - 2` for Alonzo / Mary / Allegra blocks, or `Babbage - 6` for Conway whenever they're backwards compatible.

gostkin (approved):

PR #277, opened 2023-11-21
 PlutusData cardano-node JSON encoding/decoding

rooooooooob:
Brought over from old CML's code. JSON parsing code is brought over as
well to new format.

PlutusData::BigInt -> PlutusData::Integer rename to match old CML's API
as the big part probably isn't particularly important


rooooooooob:
I made the first commit just as moving the files over from the old CML so you can diff just the other commits if you want.

SebastienGllmt (approved):

gostkin (approved):

Issue #286, opened 2023-12-26
Missing JsonSchema for AssetName


SebastienGllmt:
`AssetName` is defined in the following way

```rust
pub struct AssetName {
    pub inner: Vec<u8>,
    #[derivative(
        PartialEq = "ignore",
        Ord = "ignore",
        PartialOrd = "ignore",
        Hash = "ignore"
    )]
    #[serde(skip)]
    pub encodings: Option<AssetNameEncoding>,
}
```

This leads to the incorrectly generated JSON type

```typescript
export interface AssetNameJSON {
  inner: number[];
}
```

To avoid this issue, we usually define `JsonSchema for X` to override the generated definition, but we're missing this for `AssetName`

SebastienGllmt:
I looked through CML to see which types are effected. I feel like there are three types:

## 1. Types that have become `{ inner: ... }`
- AssetNameJSON
- DnsNameJSON
- Ipv4JSON
- Ipv6JSON
- UrlJSON
- KESSignatureJSON
- PlutusV1ScriptJSON
- PlutusV2ScriptJSON
## 2. Types that have become `[k: string]: unknown;`
- IntJSON
- ScriptJSON
- TransactionMetadatumJSON
- CredentialJSON
- NonceJSON
- PlutusDataJSON
- DatumOptionJSON
- CIP36DelegationDistributionJSON
## 3. Other
- ByronAddressJSON seems incorrect even though I remember we fixed this issue in the past (it shows the parts instead of a base58 address)

PR #287, opened 2023-12-26
Work on fixing assetname json

SebastienGllmt:
This does some of the work for #286

I didn't finish this PR, so maybe @rooooooooob you can help?

### DIFF
--- a/chain/rust/src/assets/mod.rs
+++ b/chain/rust/src/assets/mod.rs
@@ -13,7 +13,7 @@ use cml_core::error::*;
 use std::convert::TryFrom;
 
 #[derive(
-    Clone, Debug, serde::Deserialize, serde::Serialize, schemars::JsonSchema, derivative::Derivative,
+    Clone, Debug, derivative::Derivative,
 )]
 #[derivative(Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct AssetName {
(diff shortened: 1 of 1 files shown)


chain/rust/src/assets/mod.rs
```diff
@@ -13,7 +13,7 @@ use cml_core::error::*;
 use std::convert::TryFrom;
 
 #[derive(
-    Clone, Debug, serde::Deserialize, serde::Serialize, schemars::JsonSchema, derivative::Derivative,
+    Clone, Debug, derivative::Derivative,
```
SebastienGllmt, line 16: I don't remember if we need to remove `derivative::Derivative` as well

chain/rust/src/assets/mod.rs
```diff
@@ -64,3 +63,41 @@ impl From<AssetName> for Vec<u8> {
...
+        D: serde::de::Deserializer<'de>,
+    {
+        let s = <String as serde::de::Deserialize>::deserialize(deserializer)?;
+        let asset_name = hex::decode(&s).map_err(|_e| {
+            serde::de::Error::invalid_value(
+                serde::de::Unexpected::Str(&s),
+                &"invalid hex string",
+            )
+        })?;
+        // TODO: map error
```
SebastienGllmt, line 88: I don't remember how to map between these error types

SebastienGllmt:
Replaced by #300

PR #288, opened 2023-12-27
Into bytes instance for const size array-based wrapper types.

oskin1:


rooooooooob:
Should be possible to do `From` instead, see clippy lint that failed.

rooooooooob (approved):
LGTM

PR #296, opened 2024-01-08
BigInt to u128

oskin1:


rooooooooob (approved):
How far do we want to take this? I guess it's LGTM but to what point do we want to support all the explicit conversions? I guess u128 is fine. Do we want i128? I suppose those are the only two that would make sense since beyond that the are no default rust primitives, and below that there's the `Int` type.

SebastienGllmt:
Although this PR was already merged, probably this wasn't the right way to implement this. Rust comes with `From` and `TryFrom` traits for cases like this, and the only reason we had the `as_u128` pattern was because WASM needs explicit functions (which is no longer a concern with our Rust / WASM split)

This is not a big deal, but especially in cases like this where we only modified the Rust code and not the WASM code, we should probably do the more Rust-idiomatic approach

I created an issue for this: https://github.com/dcSpark/cardano-multiplatform-lib/issues/304

PR #298, opened 2024-01-23
Multi-era exts

gostkin:
Adds better support for multi-era. Mostly hash functions are added to tx bodies / blocks. Fixed pool naming as well

### DIFF
--- a/chain/rust/src/transaction/mod.rs
+++ b/chain/rust/src/transaction/mod.rs
@@ -24,7 +24,8 @@ use cbor_encodings::{
     TransactionWitnessSetEncoding,
 };
 use cml_core::ordered_hash_map::OrderedHashMap;
-use cml_core::serialization::{LenEncoding, StringEncoding};
+use cml_core::serialization::{LenEncoding, Serialize, StringEncoding};
+use cml_crypto::blake2b256;
 use std::collections::BTreeMap;
 
 #[derive(
@@ -374,6 +375,10 @@ impl TransactionBody {
             donation: None,
             encodings: None,
         }
+    }
+
+    pub fn hash(&self) -> [u8; 32] {
+        blake2b256(&self.to_cbor_bytes())
     }
 }
 
--- a/multi-era/rust/src/utils.rs
+++ b/multi-era/rust/src/utils.rs
@@ -15,6 +15,7 @@ use crate::{
     mary::MaryBlock, shelley::ShelleyBlock,
 };
 use crate::{MultiEraBlock, MultiEraTransactionBody};
+
 use cbor_event::de::Deserializer;
 use cml_chain::address::Address;
 use cml_chain::assets::{Mint, PositiveCoin};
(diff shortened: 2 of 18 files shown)


multi-era/rust/src/utils.rs
```diff
@@ -233,6 +231,24 @@ impl MultiEraBlock {
             Self::Conway(block) => block.invalid_transactions.clone(),
         }
     }
+
+    pub fn hash(&self) -> BlockHeaderHash {
+         let bytes = match self {
+             Self::Byron(block) => match block {
+                 // why there's no to_bytes or sth for byron block headers?
```
rooooooooob: @gostkin  There should be. I think it's the `ToBytes` trait? We don't implement the same traits as the other eras since we can't do encoding preservation/etc so it directly uses cbor_event's, but byron/cip25 should still have that `serialization::ToBytes` (in `core/rust/src/serialization.rs`). There's a TODO about regeneration there so we might want to think about if we want to keep this long-term or if there was a way to use the other traits for non-preserve types. For Byron it's not a big deal since everything is canonical anyway but for cip25 it's weird since some things are subsets but maybe those aren't implementing any of the traits.

rooooooooob: More context: we had these `ToBytes`/`FromBytes` traits from before we did preserve-encodings. (possibly named differently e.g.`ToFromBytes` or `ToCBORBytes/FromCBORBytes` - I don't remember). The new CML's `Deserialize`  which now has those to/from bytes methods now has the ability to preserve encodings and/or force canonical encoding which byron/cip25 don't.

Byron could kind of fake it since it's all guaranteed to be canonical anyway so both methods would do the same thing. This would have to be changed in cddl-codegen though since when you're not preserving encodings in general we don't make any assumptions about all inputs being canonical so we don't implement these traits and instead directly implement `cbor_event::Serialize` like we did before. You couldn't blanket impl it like with `ToBytes` since it would blanket impl it for everything not just the byron types.

CIP25 is weird since it's not canonical but skips some data e.g. parses a subset of objects so round-tripping isn't preserved. We could maybe regen cip25 and have it impl those types with the exception of the outermost object (the one equivalent to a tx metadatum) which IIRC doesn't implement `ToBytes` anyway.

multi-era/rust/src/utils.rs
```diff
@@ -233,6 +234,39 @@ impl MultiEraBlock {
             Self::Conway(block) => block.invalid_transactions.clone(),
         }
     }
+
+    pub fn hash(&self) -> [u8; 32] {
+        let bytes = match self {
+            Self::Byron(block) => match block {
+                // why there's no to_bytes or sth for byron block headers?
```
rooooooooob: Was this comment meant to be removed?

chain/rust/src/transaction/mod.rs
```diff
@@ -375,6 +376,10 @@ impl TransactionBody {
             encodings: None,
         }
     }
+
+    pub fn hash(&self) -> [u8; 32] {
+        blake2b256(&self.to_cbor_bytes())
```
rooooooooob, line 381: `[u8; u32]` is more general but less consistent than `TransactionHash`. I guess on the rust side there's an `Into` impl so it's okay.

rooooooooob (approved):
LGTM although I think that comment was meant to be removed? since it calls to_bytes() below it.

PR #300, opened 2024-01-26
Update JSON Schemas

### LINKED ISSUE #286: Missing JsonSchema for AssetName

rooooooooob:
Int / Ipv4 / Ipv6 / ByronAddress manually implemented Misc byte wrappers now use hex bytestring included in cddl-codegen update.
All wrappers via said cddl-codegne update now exclude the `{ inner: T }` in favor of just `T` directly. e.g. DnsName / Url even when not bytes

Fixes #286

### DIFF
--- a/specs/conway/certs.cddl
+++ b/specs/conway/certs.cddl
@@ -87,8 +87,8 @@ pool_params = ( operator:       ed25519_key_hash
               , pool_metadata:  pool_metadata / null
               )
 
-ipv4 = bytes .size 4
-ipv6 = bytes .size 16
+ipv4 = bytes .size 4 ; @custom_json
+ipv6 = bytes .size 16 ; @custom_json
 dns_name = tstr .size (0..64)
 
 single_host_addr = ( tag: 0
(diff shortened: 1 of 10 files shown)


specs/conway/certs.cddl
```diff
@@ -87,8 +87,8 @@ pool_params = ( operator:       ed25519_key_hash
               , pool_metadata:  pool_metadata / null
               )
 
-ipv4 = bytes .size 4
-ipv6 = bytes .size 16
+ipv4 = bytes .size 4 ; @custom_json
```
SebastienGllmt, line 90: Where does `@custom_json` come from? I don't see it in the cddl-codegen docs

PR #307, opened 2024-02-05
Docs Update

rooooooooob:


core/wasm/src/wasm_wrappers.rs
```diff
@@ -417,6 +417,13 @@ macro_rules! impl_wasm_cbor_api {
                 cml_core::serialization::Serialize::to_cbor_bytes(&self.0)
             }
 
+            /**
+             * Serialize this type to CBOR bytes using canonical CBOR encodings
+             */
```
rooooooooob, line 422: I added this to all the types that support preserve-encodings/etc as we expose it in rust and it can't hurt to expose it. I used it in the cbor docs example.

One small possible inconsistency is that it's not added to the byron/cip25 types as we can't guarantee anything implementing `cbor_event::Serialize` will do it canonically or not. The comment for `to_cbor_bytes()` on those types does mention that the format may or may not be canonical.

rooooooooob:
Replaced by #345 + #346

Issue #317, opened 2024-03-07
ByronBlock / MultiEraBlock explicit era tag


rooooooooob:
In Byron the era tag wrapping struct is explicitly a part of the CDDL:

```cddl
byron_block = [
     ; @name ebb_wrapper
     tag: 0, block: byron_eb_block //
     ; @name main_block_wrapper
     tag: 1, block: byron_main_block
]
```

For all the other eras there is no mention at all of this outer wrapping struct e.g. Conway (all other eras are similar):
```cddl
block =
  [ header
  , transaction_bodies         : [* transaction_body]
  , transaction_witness_sets   : [* transaction_witness_set]
  , auxiliary_data_set         : {* transaction_index => auxiliary_data }
  , invalid_transactions       : [* transaction_index ]
  ]
```

When we retrieve blocks from oura/cardano-net-fetcher we get even these blocks with explicit era tags as if it were:
```cddl
conway_block = [7, block]
```

One approach in order to respect the CDDL it could make sense to keep the tag info in the byron blocks but not touch them in the other eras, and then in `MultiEraBlock` do the check as if it were defined like:

```cddl
multi_era_block =
    byron_block            ; @name byron
  / [2, shelley_block]     ; @name Shelley
  / [3, allegra_block]     ; @name Allegra
  / [4, mary_block]        ; @name Mary
  / [5, alonzo_block]      ; @name Alonzo
  / [6, babbage_block]     ; @name Babbage
  / [7, block]             ; @name Conway
```

or similar.

We currently have `MultiEraBlock::from_explicit_network_cbor_bytes()` to parse using this era tag struct.

SebastienGllmt:
> When we retrieve blocks from oura/cardano-net-fetcher we get even these blocks with explicit era tags as if it were:

As a reminder, the Cardano network specification has a separate CDDL definition from the ledger specification and CML currently doesn't contain any codegen from the networking spec. I'm not sure the networking cddl is a particularly good state (I've never heard good things about it)

PR #318, opened 2024-03-07
MultiEraBlock::from_explicit_network_cbor_bytes EBB

rooooooooob:
Update the function to handle Byron EBBs.

We still need to figure out what to do in the general deserialize case #317 or if that behavior should just merge this function in there.

This works in the meantime and the behavior here would work regardless of how we update `ByronBlock::deserialize()`


SebastienGllmt:
Can we also add the inline comment to the hash function in this PR (or do it in a separate PR)?

SebastienGllmt (approved):

PR #321, opened 2024-03-11
TX hash mismatch for non-canonical TXs

### LINKED ISSUE #314: TransactionHash Mistmatch
I'm not sure why CML produces a different tx hash compared to Lucid CML

## CML example
```
const txhashCML = CML.hash_transaction(CML.TransactionBody.from_cbor_hex(("a500838258200d31e3060edc0422bab792b414b6920534fb61f72a24cb76c911fea670601518008258205a4a961c2ff22b870f06f5303dcfffaaf3afdb819b12f5f0e60f0540e0e7490700825820f9178e7551be72eb512f8b1b739346c84e9cda4ef8be4feada9bfff6fe47e80000018182583900bc83d1474fafc0669e360c6ef389d3874f5eff99c31f343ff2cdab0bcb68eb5507f1a2a0cb4ddb095125f0f9878568cfe2466f2969db1ed3821a02050c95a9581c748a7026de606c27c06e4b90cb2e1844b566b89fba0d282debe4b40aa14d4d794d696e746564546f6b656e01581ca11c0e63717ac9d8d7bd50c6b29111003e765d199915a02613069641a14d4d794d696e746564546f6b656e181e581ca49d85bfbd193446051e644e7ad448682a1bc1b02d5a15f78429a1f0a14d4d794d696e746564546f6b656e0a581cc7e5e285a6ca1fdd6b8b0293b4d825a9c92ae69f8fc35ae2252a697da14d4d794d696e746564546f6b656e0a581cfed7a8ee8af612b08a05304ba434ef18f37be06f144fbc190d377942a14d4d794d696e746564546f6b656e0a581c92db8682a9e00473968130a7d3ee6d1bd331812543a8584ba739c74ca14d4d794d696e746564546f6b656e01581cd3c9b677c3e599ac2eb0a455e4d5d027d5ef860caf72b5dc27ba0e16a14d4d794d696e746564546f6b656e01581ce2085cebe2ce09d4ec0cbfe965f332941686c9288199440d45e95ff0a14d4d794d696e746564546f6b656e01581c2fce0a65bac1267022f94296e3d55fb862979b13bed38521a31cd200a14d4d794d696e746564546f6b656e01021a0002fb89031a0333a0f509a1581c2fce0a65bac1267022f94296e3d55fb862979b13bed38521a31cd200a14d4d794d696e746564546f6b656e01")))
console.log("txhash CML", txhashCML.to_hex())
```
the output is the following 
`txhash CML 3ea6919295e402b9cf13afdc0f45104f9a91d299b896812f363bb8cc0f80f0a3`

## Lucid example
```
const lucidTxHash = C.hash_transaction(C.TransactionBody.from_bytes(fromHex("a500838258200d31e3060edc0422bab792b414b6920534fb61f72a24cb76c911fea670601518008258205a4a961c2ff22b870f06f5303dcfffaaf3afdb819b12f5f0e60f0540e0e7490700825820f9178e7551be72eb512f8b1b739346c84e9cda4ef8be4feada9bfff6fe47e80000018182583900bc83d1474fafc0669e360c6ef389d3874f5eff99c31f343ff2cdab0bcb68eb5507f1a2a0cb4ddb095125f0f9878568cfe2466f2969db1ed3821a02050c95a9581c748a7026de606c27c06e4b90cb2e1844b566b89fba0d282debe4b40aa14d4d794d696e746564546f6b656e01581ca11c0e63717ac9d8d7bd50c6b29111003e765d199915a02613069641a14d4d794d696e746564546f6b656e181e581ca49d85bfbd193446051e644e7ad448682a1bc1b02d5a15f78429a1f0a14d4d794d696e746564546f6b656e0a581cc7e5e285a6ca1fdd6b8b0293b4d825a9c92ae69f8fc35ae2252a697da14d4d794d696e746564546f6b656e0a581cfed7a8ee8af612b08a05304ba434ef18f37be06f144fbc190d377942a14d4d794d696e746564546f6b656e0a581c92db8682a9e00473968130a7d3ee6d1bd331812543a8584ba739c74ca14d4d794d696e746564546f6b656e01581cd3c9b677c3e599ac2eb0a455e4d5d027d5ef860caf72b5dc27ba0e16a14d4d794d696e746564546f6b656e01581ce2085cebe2ce09d4ec0cbfe965f332941686c9288199440d45e95ff0a14d4d794d696e746564546f6b656e01581c2fce0a65bac1267022f94296e3d55fb862979b13bed38521a31cd200a14d4d794d696e746564546f6b656e01021a0002fb89031a0333a0f509a1581c2fce0a65bac1267022f94296e3d55fb862979b13bed38521a31cd200a14d4d794d696e746564546f6b656e01")))
console.log("lucidTxHash", lucidTxHash.to_hex())
```

the output is the following 
`lucidTxHash 2548148ad0fb809b5756c6f8b8e1d3e705f0c4cda44af6163d6b766786d061ec`

The problem I have is when using CML, Blockfrost returns the following error
```
Error: "transaction submit error ShelleyTxValidationError ShelleyBasedEraBabbage (ApplyTxError [UtxowFailure (FromAlonzoUtxowFail (WrappedShelleyEraFailure (InvalidWitnessesUTXOW [VKey (VerKeyEd25519DSIGN \"4f26d9d8185481167e5647039d4f390c8c74a5fbca1f115709a2998e93180f39\")])))])"
```

rooooooooob:
This only concerns the free-floating `hash_transaction()`.

Fixes #314

Use `TransactionBody` instead of of [u8; 32] to be consistent with
`hash_transaction()`.

Moves all the functions that were added directly to `mod.rs` in https://github.com/dcSpark/cardano-multiplatform-lib/pull/298 files to
`utils.rs` files where they should be since they aren't auto-generated.

### DIFF
--- a/chain/rust/src/crypto/hash.rs
+++ b/chain/rust/src/crypto/hash.rs
@@ -17,7 +17,7 @@ pub fn hash_auxiliary_data(auxiliary_data: &AuxiliaryData) -> AuxiliaryDataHash 
 }
 
 pub fn hash_transaction(tx_body: &TransactionBody) -> TransactionHash {
-    TransactionHash::from(blake2b256(tx_body.to_canonical_cbor_bytes().as_ref()))
+    TransactionHash::from(blake2b256(tx_body.to_cbor_bytes().as_ref()))
 }
 
 pub fn hash_plutus_data(plutus_data: &PlutusData) -> DatumHash {
--- a/chain/rust/src/transaction/mod.rs
+++ b/chain/rust/src/transaction/mod.rs
@@ -24,8 +24,7 @@ use cbor_encodings::{
     TransactionWitnessSetEncoding,
 };
 use cml_core::ordered_hash_map::OrderedHashMap;
-use cml_core::serialization::{LenEncoding, Serialize, StringEncoding};
-use cml_crypto::blake2b256;
+use cml_core::serialization::{LenEncoding, StringEncoding};
 use std::collections::BTreeMap;
 
 #[derive(
@@ -375,10 +374,6 @@ impl TransactionBody {
             donation: None,
             encodings: None,
         }
-    }
-
-    pub fn hash(&self) -> [u8; 32] {
-        blake2b256(&self.to_cbor_bytes())
     }
 }
 
--- a/chain/rust/src/transaction/utils.rs
+++ b/chain/rust/src/transaction/utils.rs
@@ -6,9 +6,17 @@ use crate::{
     transaction::{DatumOption, ScriptRef, TransactionOutput},
     Value,
 };
-use cml_crypto::{DatumHash, Ed25519KeyHash};
+use cml_crypto::{DatumHash, Ed25519KeyHash, TransactionHash};
 
-use super::{AlonzoFormatTxOut, ConwayFormatTxOut, NativeScript, TransactionWitnessSet};
+use super::{
+    AlonzoFormatTxOut, ConwayFormatTxOut, NativeScript, TransactionBody, TransactionWitnessSet,
+};
+
+impl TransactionBody {
+    pub fn hash(&self) -> TransactionHash {
+        crate::crypto::hash::hash_transaction(self)
+    }
+}
 
 impl TransactionOutput {
     pub fn new(
--- a/multi-era/rust/src/allegra/mod.rs
+++ b/multi-era/rust/src/allegra/mod.rs
@@ -23,9 +23,8 @@ use cml_chain::transaction::{NativeScript, TransactionInput};
 use cml_chain::Withdrawals;
 use cml_chain::{DeltaCoin, LenEncoding, TransactionIndex};
 use cml_core::ordered_hash_map::OrderedHashMap;
-use cml_core::serialization::Serialize;
 use cml_core::Epoch;
-use cml_crypto::{blake2b256, Ed25519KeyHash, GenesisDelegateHash, GenesisHash, VRFKeyHash};
+use cml_crypto::{Ed25519KeyHash, GenesisDelegateHash, GenesisHash, VRFKeyHash};
 use std::collections::BTreeMap;
 
 use self::cbor_encodings::{MoveInstantaneousRewardEncoding, MoveInstantaneousRewardsCertEncoding};
@@ -187,10 +186,6 @@ impl AllegraTransactionBody {
             validity_interval_start: None,
             encodings: None,
         }
-    }
-
-    pub fn hash(&self) -> [u8; 32] {
-        blake2b256(&self.to_cbor_bytes())
     }
 }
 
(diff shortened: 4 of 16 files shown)


rooooooooob:
Why do we have this redundant functionality (both free floating `hash_*()` and `*::hash()`) for everything?

At least now one just refers to the other but it's still a bit weird. Was it easier to map a function over values from WASM or some other WASM-related ease of use? I vaguely remember hearing something about that years ago but I could be wrong. @SebastienGllmt If there's no good reason to have both I would prefer we just have one or the other to make the API simpler.

Do we want to provide some functionality for ensuring the entire tx is canonical? We could probably auto-gen or write a `make_canonical()` to structs but that could cause fees to not line up or maybe some other issues for anything in the intermediate construction that referenced hashes or sizes of components. If you're creating entirely from within CML in the first place then everything should already be canonical so maybe it's fine as is?

SebastienGllmt:
> Was it easier to map a function over values from WASM or some other WASM-related ease of use

If I remember correctly, it had to do with supporting multiple eras where the function on the type itself would always use the current era, but the standalone utility functions would be provided for old eras as well. I don't think we need this anymore with our new project configuration though

rooooooooob:
It looks like in new CML we only had the standalone and the methods were added in #298 when adding the extra multi-era functionality needed to use in Carp. I moved those to their `utils.rs` files instead of the `mod.rs` they were in.

Do we prefer the standalone or the attached methods if we pick to just have one?

What about the other hash functionality (plutus datums, metadatums).

SebastienGllmt (approved):

SebastienGllmt:
The way you did it in this PR is my preferred way 👍

SebastienGllmt:
> We could probably auto-gen or write a make_canonical()

There are some key things we shouldn't change like the tx metadata and the plutus datum. I think trying to make things canonical could make sense, but it could also introduce subtle issues and so I'm not really sure it's worth touching

rooooooooob:
I am reading CIP21, just to make sure I understand this:

>HW wallets do not serialize auxiliary data because of their complex structure. They only include the given auxiliary data hash in the transaction body.

Does this mean that the tx body has to be canonical - e.g. the hashes are canonical bytes, but were computed on arbitrary CBOR?

And the outer transaction could be encoded as arbitrary CBOR it's just the `transaction_body` that must be canonical (since that's what's signed)?

Earlier in the CIP it says:

>Transactions must be serialized in line with suggestions from [Section 3.9 of CBOR specification RFC](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9). In particular: [insert canonical CBOR description here]

which I first interpreted as the whole transaction before (thus metadatums/etc too), not just the body, but maybe it's just ambiguous wording and it just means the body?

I guess we can just keep things how they are and assume anyone implementing for HW wallets fully understands CIP21 and CBOR encodings? I'll add a line to the CBOR section of the docs stating that all structures default to canonical CBOR unless created from deserialized bytes just to make it clear.

PR #324, opened 2024-03-28
New Conway.cddl updates

rooooooooob:
Includes:
* IntersectMBO/cardano-ledger#4003
* IntersectMBO/cardano-ledger#4178
* IntersectMBO/cardano-ledger#4055
* IntersectMBO/cardano-ledger#4117
* IntersectMBO/cardano-ledger#4040

Also includes a fix for Nonce::Identity parsing introduced in #312


chain/rust/src/utils.rs
```diff
@@ -569,6 +569,302 @@ impl Deserialize for NetworkId {
...
+    }
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        Vec::<T>::json_schema(gen)
+    }
+    fn is_referenceable() -> bool {
+        Vec::<T>::is_referenceable()
+    }
+}
+
+impl<T> AsRef<[T]> for NonemptySet<T> {
```
ecioppettini, line 618: would be nice to also have `IntoIterator for &'a NonemptySet<T>` and I guess `NonemptySet<T>` I think. For loop syntax is quite convenient.

chain/rust/src/utils.rs
```diff
@@ -569,6 +569,302 @@ impl Deserialize for NetworkId {
     }
 }
 
+// Represents the cddl: #6.258([+ T]) / [* T]
+// it DOES NOT and CAN NOT have any encoding detials per element!
+// so you can NOT use it on any primitives so must be serializable directly
+#[derive(Debug, Clone)]
+pub struct NonemptySet<T> {
```
ecioppettini, line 577: does this have to be capitalized in this weird way for some reason?

rooooooooob: it's just how it is in the conway.cddl

chain/rust/src/utils.rs
rooooooooob: Just pushed this. Added:
`impl<T> IntoIterator for NonemptySet<T>`
`impl<'a, T> IntoIterator for &'a NonemptySet<T>`
`impl<'a, T> IntoIterator for &'a mut NonemptySet<T>`
+ same for `NonemptySetRawBytes`.

rooooooooob: Is there anything else we need or should `Deref` and `DerefMut` take care of that?

PR #327, opened 2024-04-30
Babbage Mint handle duplicate format

rooooooooob:
Fixes #326

PR #329, opened 2024-05-03
Fix for Babbage Mint for preprod

rooooooooob:
Fixes an issue introduced in #327

Fixes: Improper support for indefinite encoded mints + typo causing problem with multiple mints inside one policy id (not present in the test case)


SebastienGllmt (approved):
I tested on preprod and it seems the issue is fixed

PR #331, opened 2024-05-20
Add support for parsing Shelley genesis blocks

SebastienGllmt:
This is required as part of https://github.com/dcSpark/carp/pull/187

Notably, the Shelley genesis block has an `intialFunds` field and a `staking` field which essentially allows having stake pools and shelley-era keys as part of the genesis block for newly spun up networks which is useful for testing


chain/rust/src/genesis/shelley/config.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub stake: BTreeMap<Ed25519KeyHash, Ed25519KeyHash>,
+}
+
+#[derive(Debug, Clone)]
+pub struct ShelleyGenesisProtocolParameters {
+    pub a0: Fraction,
+    pub decentralisation_param: Fraction,
+    pub e_max: u64,
+    pub extra_entropy: ShelleyGenesisExtraEntropy,
+    pub key_deposit: u64,
```
rooooooooob: maybe some of these could be `Coin` when relevant? It's still just an alias but it would be consistent with the rest of the protocol param structs.

chain/rust/src/genesis/shelley/parse.rs
```diff
@@ -0,0 +1,227 @@
...
+    #[error("ParseInt: {0:?}")]
+    ParseInt(#[from] std::num::ParseIntError),
+    #[error("ParseIP: {0:?}")]
+    ParseIP(#[from] crate::certs::utils::IPStringParsingError),
+    #[error("Unexpected network type: {0:?}")]
+    ParseNetwork(String),
+}
+
+pub fn parse_genesis_data<R: Read>(json: R) -> Result<config::ShelleyGenesisData, GenesisJSONError> {
+    let data_value: serde_json::Value = serde_json::from_reader(json)?;
```
rooooooooob, line 39: Why the clone? I don't see it used anywhere else.

chain/rust/src/genesis/shelley/parse.rs
```diff
@@ -0,0 +1,227 @@
...
+                                    Some(s) => Some(Ipv6::from_str(s)?),
+                                    _ => None
+                                };
+                                relays.push(Relay::new_single_host_addr(
+                                    value.port,
+                                    ipv4,
+                                    ipv6
+                                ));
+                            },
+                            _ => panic!("Only single host address relays are supported in cardano-node Relay JSON parsing")
```
rooooooooob, line 82: If we're going to panic here it'd be good to say so in the function comment.

Edit: I read the comment on `RelayTypeMap` saying you couldn't even find JSON for other types so maybe this is okay.

rooooooooob (approved):
LGTM, just a few minor comments

chain/rust/src/genesis/shelley/raw.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ShelleyGenesisRewardAccount {
+    pub network: String,
+    pub credential: ShelleyGenesisCredential
+}
+#[allow(non_snake_case)]
+#[derive(Serialize, Deserialize, Debug)]
+pub struct ShelleyGenesisCredential {
+    // for some reason there actually is a space in the JSON key emitted by the Haskell node
+    #[serde(rename = "key hash")]
```
SebastienGllmt: It looks like cardano-node changed it so that both `key hash` and `keyHash` are accepted, so we have to handle both those possibilities here as well. Not sure if Serde has an option to enable 2 different ways to represent a field in JSON

rooooooooob: @SebastienGllmt You can use `#[serde(alias = "keyHash")]`

It can even be defined multiple times.

example:
```rust
#[derive(serde::Serialize, serde::Deserialize, Debug)]
pub struct Foo {
    #[serde(alias = "KeyHash")]
    #[serde(alias = "keyHash")]
    key_hash: String,
}

fn main() {
    let jsons = [
        "{ \"key_hash\": \"default\" }",
        "{ \"keyHash\": \"lowerCamelCase\" }",
        "{ \"KeyHash\": \"UpperCamelCase\" }",
        "{ \"keyhash\": \"should not parse\" }",
    ];
    for json in jsons {
        let foo: Result<Foo, _> = serde_json::from_str(json);
        println!("{json}      ->     {foo:?}");
    }
}
```

The first 3 are all accepted. First by it being the field name, and the next 2 for the `alias = ` marker.

Running:
```
{ "key_hash": "default" }      ->     Ok(Foo { key_hash: "default" })
{ "keyHash": "lowerCamelCase" }      ->     Ok(Foo { key_hash: "lowerCamelCase" })
{ "KeyHash": "UpperCamelCase" }      ->     Ok(Foo { key_hash: "UpperCamelCase" })
{ "keyhash": "should not parse" }      ->     Err(Error("missing field `key_hash`", line: 1, column: 33))
```

PR #332, opened 2024-06-10
feat: add native-script verify function

hadelive:


SebastienGllmt (approved):
LGTM

rooooooooob (changes requested):
Two things:

1) We put these kinds of things in `utils.rs` files instead of `mod.rs`. The `mod.rs` we try and keep as auto-generated as possible (outside of module declarations and some imports). That way it doesn't get overwritten when we need to regenerate something when they update the cddl definitions.

2) This should be in the `chain/rust` crate and then a simple binding over it for the `chian/wasm` code so that way it's exposed for rust projects too.

rooooooooob (commented):
The rust part looks good. Do we want to put the wasm bindings for that now in `chain/wasm/utils.rs`?

hadelive:
> The rust part looks good. Do we want to put the wasm bindings for that now in `chain/wasm/utils.rs`?

Yes

chain/wasm/src/transaction/utils.rs
```diff
@@ -68,6 +69,61 @@ impl NativeScript {
     pub fn hash(&self) -> ScriptHash {
         self.0.hash().into()
     }
+
+    pub fn verify(
```
rooooooooob, line 73: I meant just a wrapper around the rust code to avoid duplicating logic e.g.:
```rust
self.0.verify(lower_bound, upper_bound, key_hashes.as_ref())
```

^ untested but it's something like that

We have `AsRef` and `Into`/`From` to convert to/from the rust and WASM structs.

rooooooooob:
The checks fail with a clippy warning. Once that is fixed and the check passes I'll LGTM and merge it. Thanks.

rooooooooob (approved):

hadelive:
How long does it take for the new release?

PR #333, opened 2024-06-11
Governance Builders

### LINKED ISSUE #323: Conway Proposing/Voting RedeemerTag builder update
update builders for the two new redeemer types

rooooooooob:
Adds `VoteBuilder` and `ProposalBuilder` to facilitate creation of votes and voting proposals to transactions.

Resolves #323


rooooooooob:
I'm just going to mark this as ready for review now.  I haven't been able to test everything but I've been able to use the builders to make a few proposals (params, constitution, info) and vote on them (params, consitution). The hardfork one should be fine as there is just a version in it just like the constitution. I could test that too but I need more test ADA as the drep deposit is large. The only one I am worried about is the UpdateCommittee one but I'm not sure if I can even propose that without having any of the exisitng keys.

I only tested DRep voting as I don't have a pool and am not a part of the constitutional committee.

SebastienGllmt (approved):
approving modulo any other testing you might want to do

PR #335, opened 2024-06-12
wasm-bindgen: upgrade to 0.2.87 and remove serde-deserialize feature

ecioppettini:
Removes the deprecated serde-serialize feature (which seems unused, but I'm not a 100% sure). Also upgrades the wasm-bindgen version. I'm actually not sure if the version pinning is needed for some reason, but this is the minimum version I needed to add this as a dependency in the milkomeda bridge.

### DIFF
--- a/cml/wasm/Cargo.toml
+++ b/cml/wasm/Cargo.toml
@@ -23,4 +23,4 @@ hex = "0.4.0"
 linked-hash-map = "0.5.3"
 serde_json = "1.0.57"
 serde-wasm-bindgen = "0.4.5"
-wasm-bindgen = { version = "0.2", features=["serde-serialize"] }
+wasm-bindgen = { version = "0.2" }
--- a/cml/wasm/Cargo.toml
+++ b/cml/wasm/Cargo.toml
@@ -23,4 +23,4 @@ hex = "0.4.0"
 linked-hash-map = "0.5.3"
 serde_json = "1.0.57"
 serde-wasm-bindgen = "0.4.5"
-wasm-bindgen = { version = "0.2" }
+wasm-bindgen = { version = "=0.2.87" }
--- a/cml/wasm/Cargo.toml
+++ b/cml/wasm/Cargo.toml
@@ -23,4 +23,4 @@ hex = "0.4.0"
 linked-hash-map = "0.5.3"
 serde_json = "1.0.57"
 serde-wasm-bindgen = "0.4.5"
-wasm-bindgen = { version = "=0.2.87" }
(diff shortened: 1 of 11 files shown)


cml/wasm/Cargo.toml
```diff
@@ -23,4 +23,4 @@ hex = "0.4.0"
 linked-hash-map = "0.5.3"
 serde_json = "1.0.57"
 serde-wasm-bindgen = "0.4.5"
-wasm-bindgen = { version = "0.2", features=["serde-serialize"] }
\ No newline at end of file
+wasm-bindgen = { version = "0.2" }
```
rooooooooob: What about this version? Should we put it to `0.2.87` like all the others for consistency? Same for the multi-era crate. I think it was just on `0.2` due to copying an pasting dependencies at some point.

ecioppettini: I imagine every crate will resolve to the same version anyway. But probably yeah, if we pin one it's probably best to pin all of them.

ecioppettini: Although I'm still not really sure the pin is needed

multi-era/wasm/Cargo.toml
```diff
@@ -26,7 +26,7 @@ hex = "0.4.0"
 linked-hash-map = "0.5.3"
 serde_json = "1.0.57"
 serde-wasm-bindgen = "0.4.5"
-wasm-bindgen = { version = "0.2", features=["serde-serialize"] }
+wasm-bindgen = { version = "0.2" }
```
rooooooooob: this one too?

rooooooooob (approved):
LGTM

hadelive:
Why don't you bump to the latest version (0.2.92)?
Btw, could you merge the PR and release new version, @rooooooooob?
(We need it for this [PR](https://github.com/Anastasia-Labs/lucid-evolution/pull/144))

solidsnakedev:
please upgrade to `0.2.92` we're currently having issues with the current CML version, https://github.com/Anastasia-Labs/lucid-evolution/issues/149

ecioppettini:
> Why don't you bump to the latest version (0.2.92)?

The main reason was that `0.2.87` is the version pinned by cardano-serialization-lib, so that enables using them together. I could change this to `>=0.2.87` though.

solidsnakedev:
> > Why don't you bump to the latest version (0.2.92)?
> 
> The main reason was that `0.2.87` is the version pinned by cardano-serialization-lib, so that enables using them together. I could change this to `>=0.2.87` though.

yes, that makes sense

PR #336, opened 2024-06-18
Cost models update conway

rooooooooob:
Due to changes here: https://github.com/IntersectMBO/cardano-ledger/pull/4284
Clarification here: https://github.com/IntersectMBO/cardano-ledger/issues/4383

Allows non-Plutus languages according to the spec and changes to an indexable map type.

### DIFF
--- a/chain/rust/src/builders/tx_builder.rs
+++ b/chain/rust/src/builders/tx_builder.rs
@@ -363,7 +363,7 @@ impl TransactionBuilderConfigBuilder {
             cost_models: if self.cost_models.is_some() {
                 self.cost_models.unwrap()
             } else {
-                CostModels::new()
+                CostModels::default()
             },
             _collateral_percentage: self.collateral_percentage.ok_or(
                 TxBuilderError::UninitializedField(TxBuilderConfigField::CollateralPercentage),
--- a/chain/rust/src/builders/tx_builder.rs
+++ b/chain/rust/src/builders/tx_builder.rs
@@ -720,13 +720,12 @@ impl TransactionBuilder {
 
                 match &script_witness.script {
                     PlutusScriptWitness::Ref(ref_script) => {
-                        if self
+                        if !self
                             .witness_builders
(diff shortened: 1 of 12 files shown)


chain/rust/src/plutus/utils.rs
```diff
@@ -489,6 +495,34 @@ pub fn compute_total_ex_units(redeemers: &[LegacyRedeemer]) -> Result<ExUnits, A
     Ok(sum)
 }
 
+impl TryFrom<u64> for Language {
+    type Error = DeserializeError;
+
+    fn try_from(language: u64) -> Result<Self, Self::Error> {
```
rooooooooob, line 501: I am also thinking of just getting rid of the `Language` enum entirely and swapping it with just `pub type Language = u64` then maybe a `Languages` module with V1/V2/V3 members to avoid having to convert to/from `u64`/`Language` via `From`/`TryFrom`

chain/rust/src/plutus/utils.rs
```diff
@@ -354,52 +343,69 @@ impl CostModels {
         let mut serializer = Serializer::new_vec();
         // as canonical encodings are used, we odn't need to check the keys' bytes encodings
         // and can order this statically.
-        serializer.write_map(cbor_event::Len::Len(
```
rooooooooob, line 357: This might be the reason why the plutus stuff wasn't being accepted in March, I need to look into it more. This should be fixed now. (the length calculation didn't take into account v3)

chain/rust/src/builders/tx_builder.rs
```diff
@@ -720,13 +720,12 @@ impl TransactionBuilder {
 
                 match &script_witness.script {
                     PlutusScriptWitness::Ref(ref_script) => {
-                        if self
+                        if !self
```
SebastienGllmt, line 723: These diffs should go away if you rebase

SebastienGllmt (approved):

PR #340, opened 2024-07-06
Move RawBytesEncoding and standardize use

### LINKED ISSUE #334: nodejs: AssetName class missing `to_hex()` method
`ScriptHash` has `to_hex()` method but `AssetName` does not have this method. The closest available method is `to_cbor_hex()` which adds an extra 1-2 bytes for the CBOR encoding. `to_js_value()` returns the hex value, so this can be used as a workaround, but the defined return type for that is `any` whereas the actual return type is `string`.

rooooooooob:
Fixes #334

Moves RawBytesEncoding to cml_core::serialization as there is nothing specific about crypto with it, and it can be used in many other places (Asset names, plutus scripts, etc).

Add wasm macros for declaring the RawBytesEncoding API automatically using the rust type's trait impl

Add missing hash() to PlutusV3Script

Document AssetName on how to convert to/from using utf8 traits in rust

Use dcSpark/cddl-codegen#240 to remove confusing `get()` (what is it getting?) functions on some wrapper types. These are replaced with `RawBytesEncoding` to standardize use with the rest of CML

### DIFF
--- a/chain/rust/src/assets/mod.rs
+++ b/chain/rust/src/assets/mod.rs
@@ -12,6 +12,7 @@ use cml_core::error::*;
 
 use std::convert::TryFrom;
 
+/// Use TryFrom<&str> / TryInto<&str> for utf8 text conversion and RawBytesEncoding for direct bytes access
 #[derive(Clone, Debug, derivative::Derivative)]
 #[derivative(Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub struct AssetName {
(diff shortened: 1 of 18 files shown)


rooooooooob:
We could re-export it with a deprecated tag to avoid breaking code as well, and keep the old methods too but deprecate them in favor of the new one. Are we stable enough at this point where we should do that or is it fine as-is for breaking changes?

#339 will break API anyway necessarily anyway so that's why i didn't bother here

SebastienGllmt (approved):

PR #343, opened 2024-07-16
Remove wasm-bindgen deps from rust

rooooooooob:
Rust CML crates must use wasm-bindgen for enums as these are used from the WASM CML crates.

Using noop_proc_macro we can replace the wasm_bindgen proc macro with a no-op that does nothing when consumed directly from rust.

This is now locked behind a feature (used_from_wasm) as we can't just use cfg target checks as compiling the wasm crate (not using web-pack) would use the rust target and break.

Crates using cml-chain and cml-multi-era from a wasm crate must specify this feature to activate the wasm_bindgen dependancy instead of the no-op macro.

### DIFF
--- a/multi-era/rust/Cargo.toml
+++ b/multi-era/rust/Cargo.toml
@@ -32,7 +32,7 @@ bech32 = "0.7.2"
 hex = "0.4.0"
 
 # non-wasm
-noop_proc_macro = { version = "0.3.0", optional = true }
+noop_proc_macro = { version = "0.3.0" }
 
 # wasm
 wasm-bindgen = { version = "0.2.87", optional = true }
--- a/multi-era/rust/Cargo.toml
+++ b/multi-era/rust/Cargo.toml
@@ -13,6 +13,9 @@ keywords = ["cardano"]
 [lib]
 crate-type = ["cdylib", "rlib"]
 
+[features]
+used_from_wasm = ["wasm-bindgen"]
+
 [dependencies]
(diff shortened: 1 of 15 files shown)


multi-era/rust/Cargo.toml
```diff
@@ -23,8 +26,13 @@ derivative = "2.2.0"
...
 serde_json = "1.0.57"
 schemars = "0.8.8"
-wasm-bindgen = { version = "0.2.87" }
 
 # only for declaring hash types
 bech32 = "0.7.2"
 hex = "0.4.0"
+
+# non-wasm
+noop_proc_macro = { version = "0.3.0" }
```
rooooooooob, line 35: I had this as `optional = true` but for some reason when consuming it from dcspark-core/carp it would not be able to find the dependancy. Strangely having it as optional worked fine for the `cml-chain` dependency. Both were specified in the same way in the same `Cargo.toml` file in dcspark-core so I have  no idea why this is the case.

ecioppettini: doesn't `cml-chain` have it as `optional = false`? which would be the same as not writing it

rooooooooob: Oops, I saw it specified and I guess my brain just read it as `= true` automatically. That makes sense now.

ecioppettini (approved):
LGTM. 

Maybe the feature could be called just `wasm` or `with_wasm_enums` or something like that? But I think it's fine anyways.

PR #349, opened 2024-08-13
[WIP] Ref script size fee calculation + Ref input support (tx builder)

rooooooooob:
Fee calculation has changed to take into account ref script sizes via the new protocol parameter `min_fee_ref_script_cost_per_byte`.

TxBuilder has been updated to actually take into account reference inputs. Before it let you add them to transactions but the witness checks were oblivious to them.

Draft PR still. I have never worked with reference inputs or reference scripts before so some of this could be incorrect still


chain/rust/src/builders/tx_builder.rs
```diff
@@ -227,10 +228,54 @@ fn min_fee(tx_builder: &TransactionBuilder) -> Result<Coin, TxBuilderError> {
 fn min_fee_with_exunits(tx_builder: &TransactionBuilder) -> Result<Coin, TxBuilderError> {
```
rooooooooob, line 227: min_fee might possibly need updating I think, or even rethinking these two because we have more components to the fee calculations now - but maybe not since can you even possibly incur this ref script fee if you don't have exunits involved? but it could be involved just give temporarily incorrect values (e.g. for seeing how much doing something changes fee mid-build before plutus contract is added)

chain/rust/src/fees.rs
```diff
@@ -50,6 +62,67 @@ pub fn min_script_fee(
...
+// }
+
+/**
+ * Calculates the cost of all ref scripts
+ * * `total_ref_script_size` - Total size (original, not hashes) of all ref scripts. Duplicate scripts are counted as many times as they occur
+ */
+pub fn min_ref_script_fee(
+    tx: &Transaction,
+    linear_fee: &LinearFee,
+    total_ref_script_size: u64,
```
rooooooooob, line 69: It isn't possible to know the original sizes so we can't compute this from just the `Transaction` anymore

chain/rust/src/builders/tx_builder.rs
```diff
@@ -227,10 +228,54 @@ fn min_fee(tx_builder: &TransactionBuilder) -> Result<Coin, TxBuilderError> {
...
+    // for output in tx_builder.outputs.iter() {
+    //     if let Some(Credential::Script{ hash, .. }) = output.address().payment_cred() {
+    //         if let Some(orig_size) = ref_script_orig_sizes.get(hash) {
+    //             total_ref_script_size += *orig_size;
+    //             println!("USING REF SCRIPT {} | SIZE {}", hash.to_hex(), orig_size);
+    //         }
+    //     }
+    // }
+
+    let ref_script_orig_sizes: HashMap<ScriptHash, u64> = if let Some(ref_inputs) = &tx_builder.reference_inputs {
```
rooooooooob: From reading CIP33 it says:
```
The key idea is to use reference inputs and modified outputs which carry actual scripts ("reference scripts"), and allow such reference scripts to satisfy the script witnessing requirement for a transaction. This means that the transaction which uses the script will not need to provide it at all, so long as it referenced an output which contained the script.
```

so I'm going on the assumption that if the hash matches something created as a reference script in a reference input's output that it must be a reference script and will incur this extra fee

other option: be explicit during tx builder/input builder and get the info from there

chain/rust/src/fees.rs
```diff
@@ -50,6 +62,67 @@ pub fn min_script_fee(
...
+) -> Result<Coin, ArithmeticError> {
+    // based on:
+    // https://github.com/IntersectMBO/cardano-ledger/blob/7e65f0365eef647b9415e3fe9b3c35561761a3d5/eras/conway/impl/src/Cardano/Ledger/Conway/Tx.hs#L84
+    // https://github.com/IntersectMBO/cardano-ledger/blob/a34f878c56763d138d2203d8ba84b3af64d94fce/eras/conway/impl/src/Cardano/Ledger/Conway/UTxO.hs#L152
+    use num::{rational::Ratio, CheckedAdd, CheckedMul};
+
```
rooooooooob, line 74: we might need `Ratio<BigNum>` during calculations then conversion afterwards

chain/rust/src/builders/tx_builder.rs
```diff
@@ -227,10 +227,71 @@ fn min_fee(tx_builder: &TransactionBuilder) -> Result<Coin, TxBuilderError> {
...
+
+    println!("\n\n\n ORIG SIZES:");
+    for (hash, size) in ref_script_orig_sizes.iter() {
+        println!("  orig {} - {}", hash.to_hex(), size);
+    }
+
+    let mut total_ref_script_size = 0;
+    for utxo in tx_builder.inputs.iter() {
+        if let Some(Credential::Script { hash, .. }) = utxo.output.address().payment_cred() {
+            println!("  ------ searching: {}", hash.to_hex());
```
SebastienGllmt: Probably we want to remove all these print statements?

chain/rust/src/builders/tx_builder.rs
```diff
@@ -402,6 +463,7 @@ pub struct TransactionBuilder {
     utxos: Vec<InputBuilderResult>,
     collateral_return: Option<TransactionOutput>,
     reference_inputs: Option<Vec<TransactionUnspentOutput>>,
+    //reference_scripts_used: Vec<Script>,
```
SebastienGllmt: do we need this?

chain/rust/src/builders/tx_builder.rs
```diff
@@ -726,16 +803,27 @@ impl TransactionBuilder {
                     .for_each(|signer| self.add_required_signer(*signer));
 
                 match &script_witness.script {
-                    PlutusScriptWitness::Ref(ref_script) => {
+                    PlutusScriptWitness::Ref(ref_script_hash) => {
+                        // it could also be a reference script - check those too
+                        // TODO: do we want to change how we store ref scripts to cache the hash to avoid re-hashing (slow on wasm) every time an input is added?
```
SebastienGllmt, line 779: TODO? Okay to merge with this?

rooooooooob: With this, yes. I don't think it's too bad of a performance hit for now (did not profile it - just going off past experience with hashing in the builder and we don't support asm.js in the new CML anyway so it's not as ridiculous a difference) and it's only when using ref scripts and we can always optimize later, using this comment as a reminder.

chain/rust/src/builders/tx_builder.rs
rooooooooob: Yes. Sorry, I forgot to set it back to draft. I just wanted to check mergability.

chain/rust/src/fees.rs
rooooooooob: Done, but do we care about other usage of the fraction crate? I changed this here since it just panics even though the final operation could fall into a `u64` fine if the intermediate denom/numerators don't. The other place we use the fraction crate is in genesis parsing.

SebastienGllmt (approved):

PR #352, opened 2024-08-29
Fix BigInt => u128 conversion.

oskin1:


chain/rust/src/utils.rs
```diff
@@ -1047,6 +1047,26 @@ mod tests {
...
+    #[test]
+    fn bigint_uint_u128_roundtrip_min() {
+        let int = u64::MIN;
+        let x = BigInteger::from_int(&Int::Uint {
+            value: int,
+            encoding: None,
+        });
+        assert_eq!(x.as_u128(), Some(int as u128))
+    }
+
```
rooooooooob, line 1069: Could you add a test case where the bits wouldn't all be uniform (i.e. not all 1's)? Maybe `462_164_030_739_157_517` or something similar could do (anything where the bytes aren't interchangeable) could work. `u64::MAX` here would have all the byte chunks be 1's / `FFFF` so wouldn't this test case have passed before the change too?

rooooooooob (changes requested):
Thanks for the work! I just have the one issue with the test case so we know it's working.

chain/rust/src/utils.rs
oskin1: Done

rooooooooob (approved):

PR #357, opened 2024-09-04
PlutusData::to_cardano_node_format()

### LINKED ISSUE #356: Non-canonical CBOR Serialization (Optional)
Tools like Aiken and Plutarch add an extra underscore `_` to items during CBOR serialization. 

This inconsistency becomes problematic when serialized data is hashed, as the resulting hash differs between tools. 

In Aiken, the below serialization will produce a CBOR value like `[ _ <hashed_value>]`, whereas CML's serialization would produce `[ <hashed_value>]` without the underscore.
Example:
```
  let datum_tag =
          out_ref
            |> serialise_data
            |> blake2b_256
```

Even though CML follows [RFC-7049-section3.9](https://datatracker.ietf.org/doc/html/rfc7049#section-3.9) The issue has resulted in some developers discarding CML in favour of other tools that maintain the non-canonical serialization.

CML should offer an option to choose between canonical and non-canonical serialization to allow compatibility across different tools.

rooooooooob:
Returns the same datum but with the encoding details changed so that it will match cardano-node/CSL/Lucid

This is only in cases where dealing in raw bytes and/or `from_cbor_bytes()` is somehow not possible, as that solution is 100% fool-proof with any tool always for calculating hashes.

Fixes #356

### DIFF
--- a/chain/rust/src/plutus/utils.rs
+++ b/chain/rust/src/plutus/utils.rs
@@ -5,6 +5,7 @@ use crate::json::plutus_datums::{
     decode_plutus_datum_to_json_value, encode_json_value_to_plutus_datum,
     CardanoNodePlutusDatumSchema,
 };
+use crate::utils::BigInteger;
 use cbor_event::de::Deserializer;
 use cbor_event::se::Serializer;
 use cml_core::error::*;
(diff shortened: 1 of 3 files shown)


SebastienGllmt (approved):

SebastienGllmt:
Perhaps this should be mentioned in the docs if we think this is a common issue?

rooooooooob:
@SebastienGllmt I'm not sure if we should be encouraging this in situations where it's avoidable. It's not good practice to hope that every little bit will line up when creating datums from scratch using a specific tool. It's much better to use `PlutusData::from_cbor_bytes()` on the bytes created with whatever other tool/on-chain data you're working with and know for sure you won't run into issues regardless of the tool. We already have a CBOR section in the docs as of #345. I could expand upon it if needed or put even more emphasis on `from_cbor_bytes()`. It already mentions how it can make the hash change if 1 thing is encoded differently and that using types created from the cbor bytes using CML avoids these issues.

Issue #354, opened 2024-09-02
Possibly incorrect min_fee calculation.


Bromel777:
Hi, I'm trying to calculate min_fee for this tx by calling `min_fee(true)`, the result is less than real.
[...]

PR #359, opened 2024-09-10
Ref script fee fixes

rooooooooob:
* Check everywhere that could use a ref script (not sure if it's allowed in all these spots)

* Duplicate size for different ref inputs with the same ref script as per Haskell code

* look in spent inputs too (as this is possible to ref a script from)

Possibly addresses #354


SebastienGllmt (approved):

SebastienGllmt (commented):
Are there any test vectors for this?

rooooooooob:
I'm going to wait at least a little bit until we can see if #354 is fixed by this before merging. If the guy who made the issue responds with what the actual inputs were (he only posted final cbor so any utxo/orig script info isn't preserved) then we can make test vectors out of that.

PR #361, opened 2024-09-20
shelley genesis: support fractional slot length

ecioppettini:
when trying to parse the shelley genesis file generated by yaci-devkit I got an error because the slotLength it's `1.0` instead of `1`.  It seems like the value can be actually have decimals (e.g https://github.com/bloxbean/yaci-devkit/blob/bef2cf9252972b6935b59650b347d95d78480278/applications/cli/src/main/java/com/bloxbean/cardano/yacicli/localcluster/ClusterCommands.java#L98 ), although I'm actually not sure if there is a spec for these fields anywhere.


chain/rust/src/genesis/shelley/config.rs
```diff
@@ -18,7 +18,7 @@ pub struct ShelleyGenesisData {
     pub network_magic: u64,
     pub protocol_params: ShelleyGenesisProtocolParameters,
     pub security_param: u64,
-    pub slot_length: u64,
+    pub slot_length: String,
```
rooooooooob: Shouldn't this be a `Fraction` like `active_slots_coeff`? That one is only a `String` to avoid `f64` losing precision in `raw.rs`'s field, but is `Fraction` here.

rooooooooob (changes requested):

SebastienGllmt (approved):

rooooooooob (approved):
LGTM

Issue #364, opened 2024-11-21
`258` tag for sets is not being correctly serialized


SeungheonOh:
When serializing `set` and `nonempty_set`, current implementation of serializers does not correctly include `258` tag that will be required starting Polmin hardfork. 

I believe @solidsnakedev have working solution for this issue. This issue is for tracking purposes.

rooooooooob:
The reason we don't serialized this way **_as the default behavior_**  is to make txs smaller, as it's not currently required.

>current implementation of serializers does not correctly include 258 tag 

It **_is_** correctly serialized for any transactions that were created using that format. It will be required later as you have mentioned, but when that happens there might be other changes to the binary spec and we want CML to properly support exactly that hardfork which would require removing `NonEmptySet` 100% from the default crate and only using it in the `cml-multi-era/src/babbage` module. We have an approach of not mixing eras and the `cml-chain` crate only supports the current era, whereas previous eras go in `cml-multi-era`.

SeungheonOh:
I see. I agree that this is the sensible default behavior for mainnet. However, Polmin fork have arrived in preview and testnet, and since this tag is required for next hardfork I'm seeing some transactions failing. 

Perhaps we can have "prerelease" version of CML that supports new serialization format to test before hardfork arrives in mainnet?

rooooooooob:
I checked the binary specs folder in the ledger specs and I don't see a .cddl definition yet for Polmin. If this is the only binary spec change I am fine with making the default behavior to be to serialize the tag, but we must still respect the encodings for current Conway/chang 1 transactions so that round-tripping works.

SeungheonOh:
As far as I can tell, they will be using same conway format, but following comment states tags for set will be enforced starting next hardfork.

https://github.com/IntersectMBO/cardano-ledger/blob/242547effb1eb3e1e710f1565cbb82ff676df3b5/eras/conway/impl/cddl-files/extra.cddl#L3

I'm not knowledgeable enough on how binary format is dealt with on ledger and everything to insist on one thing over another, but having binary tag feels like a sensible update from what I can tell from the comment

solidsnakedev:
This is the workaround we need, but as you said we need to support mainnet as well ,is there a possibility to make this tag serializtion optional, so we can switch on and off between tesnet and mainnet?

```rust
impl<T> From<Vec<T>> for NonemptySet<T> {
    fn from(elems: Vec<T>) -> Self {
        Self {
            elems,
[...]

PR #365, opened 2024-11-22
fix: use 258 tag for sets

hadelive:
#364


rooooooooob (approved):
LGTM, the failing lint isn't related to this code, it must have been an update to cargo adding new lints

PR #372, opened 2025-01-24
add wasm functions to get txid for genesis utxos

SebastienGllmt:
I need these utility functions to easily query utxorpc on genesis utxo data

### DIFF
--- a/chain/wasm/src/genesis/network_info.rs
+++ b/chain/wasm/src/genesis/network_info.rs
@@ -1,6 +1,7 @@
 use crate::byron::utils::ProtocolMagic;
 use cml_core_wasm::impl_wasm_conversions;
 use wasm_bindgen::prelude::wasm_bindgen;
+use cml_crypto_wasm::TransactionHash;
 
 #[wasm_bindgen]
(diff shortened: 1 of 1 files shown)


chain/wasm/src/genesis/network_info.rs
```diff
@@ -43,3 +44,36 @@ impl NetworkInfo {
...
+    let redeem = cml_chain::genesis::byron::parse::redeem_pubkey_to_txid(
+        &pubkey.clone().as_ref().0, 
+        protocol_magic.map(|pm| ProtocolMagic::new(pm).into())
+    );
+    ByronGenesisRedeem(redeem)
+}
+
+#[wasm_bindgen]
+pub fn genesis_txid_shelley(address: &crate::address::Address) -> TransactionHash {
+    TransactionHash::from(cml_chain::genesis::shelley::parse::redeem_address_to_txid(&address.clone().into()))
```
rooooooooob, line 78: I think you can do `address.as_ref()` and avoid the `clone()`/`into()`.

rooooooooob (commented):
LGTM besides the one small comment but it's already been merged so it's not really worth changing as its just a minor overhead

PR #376, opened 2026-05-29
Fix reference script fee calculation

colll78:
This PR updates min_fee_with_exunits to match Conway ledger reference-script fee accounting.

Previously, CML only counted reference-script bytes when a reference input’s script hash matched a spent script-address input. That missed cases where a transaction spends a regular input carrying a reference script, and it also did not match the ledger’s actual txNonDistinctRefScriptsSize behavior.

Changes:

Count reference scripts from regular inputs ∪ reference inputs.
De-duplicate by TransactionInput, so the same TxIn appearing in both sets counts once.
Count the same script multiple times when it appears on different UTxOs, matching ledger behavior.
Handle native reference scripts without panicking.
Change reference-script tier fee rounding from ceil to floor, matching Conway tierRefScriptFee.

### DIFF
--- a/chain/rust/src/builders/tx_builder.rs
+++ b/chain/rust/src/builders/tx_builder.rs
@@ -1702,7 +1702,7 @@ pub fn add_change_if_needed(
     let fee = match &builder.fee {
         None => builder.min_fee(include_exunits),
         // generating the change output involves changing the fee
-        Some(_x) => return Ok(false),
+        Some(set_fee) => Ok(set_fee.clone()),
     }?;
 
     // note: can't add datum / script_ref to change
--- a/chain/rust/src/builders/tx_builder.rs
+++ b/chain/rust/src/builders/tx_builder.rs
@@ -18,7 +18,7 @@ use crate::assets::MultiAsset;
 use crate::assets::{AssetArithmeticError, Mint};
 use crate::auxdata::AuxiliaryData;
 use crate::builders::output_builder::TransactionOutputBuilder;
-use crate::certs::{Certificate, Credential};
+use crate::certs::Certificate;
 use crate::crypto::hash::{calc_script_data_hash, hash_auxiliary_data, ScriptDataHashError};
 use crate::crypto::{BootstrapWitness, Vkeywitness};
 use crate::deposit::{internal_get_deposit, internal_get_implicit_input};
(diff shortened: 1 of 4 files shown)


scripts/publish-helper.js
```diff
@@ -10,7 +10,7 @@ const pathToRepo = path.join(__dirname, '..', crateName, 'wasm');
 const oldPkg = require(`${pathToRepo}/publish/package.json`);
 
 const packageNameRoot = hyphenRepoName.split("-wasm")[0];
-oldPkg.name = '@dcspark/' + packageNameRoot + buildType;
```
SebastienGllmt, line 13: you'll need to revert this one

chain/rust/src/builders/tx_builder.rs
```diff
@@ -224,41 +226,42 @@ fn min_fee(tx_builder: &TransactionBuilder) -> Result<Coin, TxBuilderError> {
...
-    let full_tx = fake_full_tx(tx_builder, tx_builder.build_body()?)?;
-    // we can't know the of scripts yet as they can't be calculated until we build the tx
-
-    fn ref_script_orig_size_builder(utxo: &TransactionUnspentOutput) -> Option<(ScriptHash, u64)> {
-        utxo.output.script_ref().map(|script_ref| {
-            (
-                script_ref.hash(),
-                script_ref
-                    .raw_plutus_bytes()
-                    .expect("TODO: handle this")
```
SebastienGllmt, line 237: TODO?

chain/rust/src/builders/tx_builder.rs
```diff
@@ -1718,7 +1721,7 @@ pub fn add_change_if_needed(
     let fee = match &builder.fee {
         None => builder.min_fee(include_exunits),
         // generating the change output involves changing the fee
-        Some(_x) => return Ok(false),
+        Some(set_fee) => Ok(set_fee.clone()),
```
SebastienGllmt, line 1724: I think this now contradicts the comment directly above it?

chain/wasm/src/lib.rs
```diff
@@ -1025,7 +1025,28 @@ impl_wasm_list!(
     StakeCredentialList
 );
 
-pub type SubCoin = Rational;
+#[derive(Clone, Debug)]
+#[wasm_bindgen]
+pub struct SubCoin(cml_chain::SubCoin);
```
SebastienGllmt, line 1030: Is this needed for anything?

chain/rust/src/builders/tx_builder.rs
colll78: The TODO is what was already there prior to any of my changes.

chain/rust/src/builders/tx_builder.rs
colll78: The comment was already contradicted, generating the change output did not change the fee if there was a fee set.

